Issue #105, opened 2014-03-12
support for protobuf


emicklei:
Somehow find a way to add a codec for application/x-google-protobuf such that Read and WriteEntity work with this binary format.

PR #128, opened 2014-06-17
Introducing the EntityEncoder type, to handle different entities

wongak:
In preparation for #105 (which might also be of interest for me), I'd suggest to create an Interface, which describes how Entities should be handled.

What do you think?

### DIFF
--- a/entity.go
+++ b/entity.go
@@ -33,8 +33,11 @@ type EntityEncoder interface {
 	// MIME must return the MIME type to be used in the Content-Type header
 	MIME() string
 	// SetRequest will receive the current request
-	// It must only operate on instances created with new
+	// It must only operate on instances created with New
 	SetRequest(*Request)
+	//  SetResponse will receive the current response
+	// It must only operate on instances created with New
+	SetResponse(*Response)
 	// Marshal to be used in Response writing
 	Marshal(v interface{}) ([]byte, error)
 	// MarshalIndent for PrettyPrint cases
@@ -92,6 +95,9 @@ func (e *JSONEntity) MIME() string {
 // no-op
 func (e *JSONEntity) SetRequest(r *Request) {}
 
+// SetResponse is a no-op
+func (e *JSONEntity) SetResponse(r *Response) {}
+
 // Marshal passthrough json.Marshal
 func (e *JSONEntity) Marshal(v interface{}) ([]byte, error) {
 	return json.Marshal(v)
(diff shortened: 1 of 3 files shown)


mc-team:
are you planing to merge this request ?

emicklei:
Yes, I intend to add this feature; it needs some work as it cannot be merged as is.

emicklei:
Just to let any watchers know that I am working on getting feature in. The main change I am hoping to achieve is that each Encoder and Decoder (separate them?) is stateless.

jchenry:
+1

emicklei:
package restful
		
		import "encoding/json"
		
		type EntityReader interface {
			// Read a serialized version of the value from the request.
			// The Request may have a decompressing reader.
			Read(req *Request, v interface{}) error
		}
		
		type EntityWriter interface {
			// Write an serialized version of the value on the response.
			// The Response may have a compressing writer.
			Write(req *Request, resp *Response, v interface{}) error
		}
		
		type entityJSON struct{}
		
		// Read unmarshalls the value from JSON
		func (e entityJSON) Read(req *Request, v interface{}) error {
			decoder := json.NewDecoder(req.Request.Body)
			decoder.UseNumber()
			return decoder.Decode(v)
		}
		
		// Write marshalls the value to JSON and set the Content-Type Header.
		func (e entityJSON) Write(req *Request, resp *Response, v interface{}) error {
			if v == nil {
				// do not write a nil representation
				return nil
			}
			if resp.prettyPrint {
				// pretty output must be created and written explicitly
				output, err := json.MarshalIndent(v, " ", " ")
				if err != nil {
					return err
				}
				resp.Header().Set(HEADER_ContentType, MIME_JSON)
				_, err = resp.Write(output)
				return err
			}
			// not-so-pretty
			resp.Header().Set(HEADER_ContentType, MIME_JSON)
			return json.NewEncoder(resp).Encode(v)
		}

emicklei:
This is taken from an older local branch. 

The idea is to have 2 interfaces allowing different structs to implement the Read and Write. The Read requires a Request and a value to populate. If it fails it returns with an error. Write an error response will happen in the container.

The Write requires both a Request and a Response with the value. The writing implementation may need additional information from the Request. The example implementation here reflects that of the current one in Response. (that's why there is pretty handling).

Just like the proposal for @wongak , there needs to be a way to register readers and writers (i call them like this because the R,R has a ReadEntity and WriteEntity respectively). 

Finding out what MIME to read is done in the Request. The MIME to write in the Response which is determined by looking at the Header and Defaults.

Although I did not write the XML version, I guess that it will be possible to define one with this interface. The bigger question is whether these interface suffice for the encodings you have in mind (application/x-protobuf, text/csv, etc.).

Please comment on this proposal.

stuart-warren:
Looks alright to me :+1:

emicklei:
i just push the branch for this feature. https://github.com/emicklei/go-restful/tree/entityplug

emicklei:
Please verify the current implementation in the entityplug branch. I am about to merge this into the master.

emicklei:
The entityplug branch has been merged into the branch.

PR #173, opened 2015-01-12
Swagger items should be object

m0sth8:
According to the documentation https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#434-items-object Items field should be object, not an array.

Also this PR fixes bug with slices, because []Address is equal []*Address in swagger representation


emicklei:
Thank you for reporting and fixing this. I will review the changes and do the merge today.

m0sth8:
Also, I noticed that sometimes tests fail. I suppose it's because of random ordering of maps since go 1.3

emicklei:
Yes, it is a problem I kind of ignored for a while. I think it can be fixed using DeepEquals on the unmarshalled expected value.

PR #175, opened 2015-01-14
change Container.Handle function to be filtered before handler process

nullbus:
Currently Conatiner's Handle() function does not take any container's filtering process, so I made filtering wrapper for this.
I think it must be provided for some situation such as using http.FileServer for authorized user.


nullbus:
Yes, exactly.

emicklei:
Ok. As this changes the api implementation of the package, I propose that we instead add a function e.g. HandleFiltered (pattern,handler)

nullbus:
Oh this implementation should be. Thanks!

PR #184, opened 2015-03-09
In routefunction_test, use Fatalf instead of Logf.

geowa4:
Logf output is only shown on test failure. With this change, the unit test now fails, and I'm not sure why.


emicklei:
Good catch. Will have a look the test as well.

emicklei:
Yes, this is documented , see: func (r *Response) WriteHeader(httpStatus int)
For the purpose of this, I changed it to use code 204.  I need to think about if 404 must be handled differently

PR #188, opened 2015-03-16
Use an interface for Logger instead of concrete type

### LINKED ISSUE #187: TraceLogger should accept an interface, not a type
There are plenty of logging package altenatives to the standard library log, and at least some of them try to be (somewhat) API-compatible with it. I'm using the popular [logrus](https://github.com/Sirupsen/logrus) which should be compatible to the extent required by go-restful, but the fact that `func TraceLogger(logger *log.Logger)` specifies a concrete type rather then an interface means I must use the standard library all over, or have two logging packages present in my code.

The [godoc](http://golang.org/pkg/log/) doesn't define an interface for `log`, somewhat surprisingly, but it seems like a natural thing to want to define. go-restful could define its own logger interface (with the subset of the logging API it used), which would also allow wrappers to be written for whatever logging package one uses, even if it uses its own API.

If this is acceptable, I could try to put together a PR. Would you take one?

GuyMorgan:
As easy as that. Not really sure how to test this, I guess the type checker does that for us.
I did a quick test with log.Logger and logrus.logger and both work, tests pass as well.

closes #187

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -1,16 +1,32 @@
 package restful
 
-import "log"
-
 // Copyright 2014 Ernest Micklei. All rights reserved.
 // Use of this source code is governed by a license
 // that can be found in the LICENSE file.
 
 var trace bool = false
-var traceLogger *log.Logger
+var traceLogger Logger
+
+// Logger corresponds to a subset of the interface satisfied by stdlib log.Logger
+type Logger interface {
+	Fatal(v ...interface{})
+	Fatalf(format string, v ...interface{})
+	Fatalln(v ...interface{})
+	// Flags() int
+	// Output(calldepth int, s string) error
+	// Panic(v ...interface{})
+	// Panicf(format string, v ...interface{})
+	// Panicln(v ...interface{})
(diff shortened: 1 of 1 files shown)


GuyMorgan:
Actually only Print(f). But, assuming users implement a wrapper to conform to this interface, if you ever need another method from the stdlib interface you would not be able to include it without breaking their build. seemed prudent to include everything go-restful will ever need.

That said, I can trim it to the minimum if you prefer since  it also has the downside of requiring lots of boilerplate.

emicklei:
Ok, I understand the reason now. 

It is indeed a tradeoff, now loggers need to implement those as well.

As for the panics, the effective go page mentions: ```real library functions should avoid panic.```
I think we can do without them. Developers can still choose to do a Print and call panic explicitly.

func (*Logger) Panic says:
```Panic is equivalent to l.Print() followed by a call to panic().```

GuyMorgan:
Done.

emicklei:
Thank you for the discussion and the resulting improvements.

Issue #170, opened 2014-12-03
Router does not use Content-Type if Content-Length not supplied


dkolbly:
I have a client that is sending data using `multipart/form-data` content type and is not supplying a content-length because that is not easily determinable.  My go-restful server defines two handlers for that path, one with a `Consumes("application/json")` and one with a `Consumes("multipart/form-data")`.  But go-restful does not appear to be respecting the consumes clause and picks the first defined route despite the correct specification of the content-type.

I believe [this line](https://github.com/emicklei/go-restful/blob/master/jsr311.go#L60) in `jsr311.go` is causing the problem, since in this case the `httpRequest.ContentLength` is `-1` and so it is skipping the check by contentType.

I have a patch that changes it to `httpRequest.ContentLength != 0` but I'm not sure if that's the right fix or if I'm missing something else.

nikhiljindal:
From https://jsr311.java.net/nonav/releases/1.1/spec/spec3.html, "Obtain the media type of the request. If the request does not contain a Content-Type header then use application/octet-stream"
Server should always do the content-type check. "application/octet-stream" should be used as the default value if the request does not have a content-type. Hence the current code and the proposed fix both seem wrong.

Here is what we can do:
contentType := httpRequest.Header.Get(HEADER_ContentType)
if httpRequest.ContentLength == 0 {
  contentType = "application/octet-stream"
}
// Compare contentType with Content-Type accepted by the server.

emicklei:
I wonder whether this behavior is specific to jsr311, i.e. will it become router impl specific in go-restful.

emicklei:
@nikhiljindal , please verify the change. Feel free to reopen if problems occur

wuub:
@emicklei AFAICT this change introduced a very serious regression for GET requests.
A standard GET looks like this:
```
GET / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, compress
Host: [Host]
User-Agent: [U-A]
```

No Content-Type and no Content-Lenght, as expected. In such case go-restful sets the Content-Type to `application/octet-stream`, which at least in theory is RFC compliant. But then it tries to validate this Content-Type against a list specified in .Consumes(), which causes GETs to error with 415 in almost all non-trivial cases.

I'm pretty confident that requests with empty body should not be checked against .Consumes() list,  and GET/HEAD/OPTIONS  should probably  not be checked against .Consumes()  at all.

What do you think?

wuub:
[...]

PR #195, opened 2015-03-23
Fix 415 HTTP errors for idempotent requests introduced in #170

wuub:
see #170 for comments

### DIFF
--- a/route.go
+++ b/route.go
@@ -90,6 +90,13 @@ func (r Route) matchesAccept(mimeTypesWithQuality string) bool {
 
 // Return whether the mimeTypes match to what this Route can consume.
 func (r Route) matchesContentType(mimeTypes string) bool {
+
+	// idempotent methods with empty content always match Content-Type
+	m := r.Method
+	if m == "GET" || m == "HEAD" || m == "OPTIONS" || m == "DELETE" {
+		return true
+	}
+
 	// check for both defaults
 	if len(r.Consumes) == 0 && mimeTypes == MIME_OCTET {
 		return true
--- a/web_service_test.go
+++ b/web_service_test.go
@@ -85,6 +85,17 @@ func TestSelectedRoutePath_Issue100(t *testing.T) {
 	}
 }
 
+func TestContentType415_Issue170(t *testing.T) {
+	tearDown()
+	Add(newGetOnlyJsonOnlyService())
+	httpRequest, _ := http.NewRequest("GET", "http://here.com/get", nil)
+	httpWriter := httptest.NewRecorder()
+	DefaultContainer.dispatch(httpWriter, httpRequest)
+	if 200 != httpWriter.Code {
+		t.Errorf("Expected 200, got %d", httpWriter.Code)
+	}
+}
+
 func newPanicingService() *WebService {
 	ws := new(WebService).Path("")
 	ws.Route(ws.GET("/fire").To(doPanic))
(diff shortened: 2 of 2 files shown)


wuub:
sorry for the accidental force push :(

emicklei:
So if the request normally does not contain a body (entity) then the content-type check can be ommitted. 

I tried confirming this for the methods listed in the condition by interpreting the http spec. So did others http://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request

I am eager to fix this issue and perhaps this fix will help us for 90% of the cases. What if somebody defines 2 GET routes consuming two different content-types? With this fix, the first wins.

@dkolbly mentioned in Issue #170 that simply testing the CT header is not enough. But again, who is going to stream data using the idempotent methods.

wuub:
I'm not 100% sure this is the best way to fix this in go-restful, the problem is tricky and caused by several interlocking behaviours that from my point of view are also a bit suspicious.  

>  What if somebody defines 2 GET routes consuming two different content-types? With this fix, the first wins.

Probably. But registering two GET endpoints with different consuming types is in itself a strange behaviour since almost no clients send Content-Type with get requests.  Different Accept: headers, should be handled correctly, sure, different Consumes, I'm not convinced :) 


TBH I think the best behaviour here should be:
a) not setting default Content-Type when client sent an empty body
b) treating `nil` Content-Type as an indication of empty body (entity) and allowing it implicitly 
c) always validating client-provided Content-Type, event if it's attached to an idempotent method with empty body (entity)


But such logic was difficult to implement, since I'm not very familiar with go-restful internals.

dkolbly:
Yeah, my case was streaming `POST`.  I agree on GET having two routes with different Consumes seems a bit weird.

emicklei:
new proposal:

```
// Return whether this Route can consume content with a type specified by mimeTypes (can be empty).
func (r Route) matchesContentType(mimeTypes string) bool {

	if len(r.Consumes) == 0 {
		// did not specify what it can consume ;  any media type (“*/*”) is assumed
		return true
	}

	if len(mimeTypes) == 0 {
		// idempotent methods with (most-likely or garanteed) empty content match missing Content-Type
		m := r.Method
		if m == "GET" || m == "HEAD" || m == "OPTIONS" || m == "DELETE" || m == "TRACE" {
			return true
		}
		// proceed with default
		mimeTypes = MIME_OCTET
	}
```

1) According to JSR311 (3.5) if Consumes or Produces is empty then process it like it was `*/*`
2) It adds the special case where the ContentType is missing (is empty). Then the idempotent method check is done.

@wuub 
a) there is no way to tell whether the body is empty (streaming)
b) no Content-Type means application/octet-stream. For certain methods the match process is continued.
c) realized

About the go-internals, the Route.matchesContentType is part of the candidate selection of Routes which also includes matching the Produces "annotation" with the Accept header.

I still feel that this is a compromise ; the idempotent method check allows clients to send requests with arbitrary data as long as they "forget" to send the Content-Type (and have a matching Produces).
This check is needed to satisfy e.g. testcase TestContentType415_Issue170

wuub:
AFAICT  `if len(mimeTypes) == 0 {` wil never be true, unless you change https://github.com/emicklei/go-restful/blob/master/jsr311.go#L65 here

emicklei:
yes , that is part of the proposal

emicklei:
Agree, the request will probably provide some hints via Headers whether a body can be expected. I will merge my proposal and try to refine the solution

PR #197, opened 2015-03-26
Fix to match any mime types when Route produce "*/*"

superbrothers:
This PR fix to match any mime types when Route produce "\*/\*" (such as proxy).

```go
	proxyRoute := ws.Method(method).Path(path).To(routeFunction(proxyHandler)).
		Filter(monitorFilter("PROXY", resource)).
		Doc("proxy " + method + " requests to " + kind).
		Operation("proxy" + method + kind).
		Produces("*/*").
		Consumes("*/*")
	addParams(proxyRoute, params)
	ws.Route(proxyRoute)
```
https://github.com/GoogleCloudPlatform/kubernetes/blob/cfb6f1199bae1c0c7968c7322dd1f0037624236c/pkg/apiserver/api_installer.go#L638

```
% curl -H "Accept: application/json" https://example.com/...
// 406: Not Acceptable
```

### DIFF
--- a/route.go
+++ b/route.go
@@ -79,8 +79,8 @@ func (r Route) matchesAccept(mimeTypesWithQuality string) bool {
 		if withoutQuality == "*/*" {
 			return true
 		}
-		for _, other := range r.Produces {
-			if other == withoutQuality {
+		for _, producibleType := range r.Produces {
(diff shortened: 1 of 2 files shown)


emicklei:
Thanks @superbrothers

PR #198, opened 2015-03-26
fixes #196 Set description property on model fields

### LINKED ISSUE #196: Set description property on model fields
I'm interested in having better API documentation.  I'm inclined to believe that the addition of a new structure tag, say "restful" (as a peer to the json tag) to structure members may allow automatic generation of richer documentation for structure properties in the models.

Alternatively, having a documented API to set this would work, but that would be pretty awkward.  Right now the API is fairly automatic, so it would be nice to preserve that.

I can probably code this up sometime this week if there is interest.

gdamore:
This creates the following structure tags to control the model properties:

swagger%description -- string description
swagger%type -- string "type"  (e.g. "number")
swagger%enum -- string formatted array of enumeration values -- members separated by "|" characters
swagger%default -- string containing value of "defaultValue"
swagger%minimum, swagger%maximum -- what you think (string)
swagger%unique -- either "true" or "false" setting of uniqueItems

### DIFF
--- a/swagger/model_builder.go
+++ b/swagger/model_builder.go
@@ -44,9 +44,11 @@ func (b modelBuilder) addModel(st reflect.Type, nameOverride string) {
 		field := st.Field(i)
 		jsonName, prop := b.buildProperty(field, &sm, modelName)
 		if descTag := field.Tag.Get("description"); descTag != "" {
-			prop.Description = descTag
+			if prop.Description == "" {
+				prop.Description = descTag
+			}
 		}
-		// add if not ommitted
+		// add if not omitted
 		if len(jsonName) != 0 {
 			// update Required
 			if b.isPropertyRequired(field) {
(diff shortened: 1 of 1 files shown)


emicklei:
This proposal seems to use a non-conventional notation for tag use. Why did you not choose it like this:
```
type Person struct {
     Name   string  `swagger:"description=Full name"`
     Age      int   `swagger:"minimum=1,maximum=120"`
}
```

Btw, the package already support `description` as a tag.

gdamore:
Because description could have embedded commas.  The easiest way to avoid creating limitations on the tag contents was to use a different tag for each field. 

Does description actually work?  I can use that tag name, I wasn't aware of its existence. I qualified the tags with swagger to avoid conflicts.

emicklei:
Please have a look at Issue #202 and its extension

bgrant0607:
We are using description tags.

Why not just use all separate tags? They are useful independent of swagger.
```
type Person struct {
     Name   string  `description:"Full name"`
     Age      int   `minimum:"1" maximum:"120"`
}
```

gdamore:
I agree. I will resubmit with those changes.  At the time I wanted to avoid conflicting usage in the namespace but it seems unlikely to be a problem. 

I'm on vacation right now without good internet access but I will resubmit in a week after  I have returned.

jimmidyson:
@bgrant0607 @gdamore I'd like to see support for JSON schema validation constraints in tags, see http://json-schema.org/latest/json-schema-validation.html.

In particular the following would be super useful:

Strings: maxLength, minLength, pattern
Numbers: multipleOf, maximum, minimum
Arrays: maxItems, minItems, uniqueItems

And enum for any types.

Of course the wider support the better!

bgrant0607:
Is this being actively worked on?

emicklei:
Unless @gdamore is ready to update the PR, I will work and probably finish this today.

swagger/model_builder.go
```diff
@@ -71,6 +73,67 @@ func (b modelBuilder) isPropertyRequired(field reflect.StructField) bool {
...
+	}
+}
+
+func (prop *ModelProperty) setProps(field reflect.StructField) {
+	prop.setType(field)
+	prop.setDescription(field)
+	prop.setEnumValues(field)
+	prop.setMinimum(field)
+	prop.setMaximum(field)
+	prop.setMaximum(field)
```
jimmidyson: Duplicate line?

emicklei:
did you load  https://github.com/emicklei/go-restful/commit/f949206d3742e97d3f4344e2db0ef97ebc28bafc ?

emicklei:
o, ic, yes

PR #199, opened 2015-03-27
Use "string" for things of kind "string"

brendandburns:
User "integer" for things of kind "uint64"


brendandburns:
also use 'type': 'any' for generic maps

PR #200, opened 2015-03-27
Fix up support for anonymous fields.

brendandburns:
Properly handle anonymous fields that are structs, by adding the model to the global model list.


emicklei:
thank you for contributing again

PR #203, opened 2015-04-11
Embedded structs can have a JSON tag for nesting

smarterclayton:
A nested embedded struct (embedded in Golang, nested in JSON) is possible by
default. The inline tag forces a struct to be included by default. This
commit adds that check.

@bgrant0607 FYI we need this for v1beta3 metadata


bgrant0607:
Thanks.
cc @brendandburns @nikhiljindal

PR #204, opened 2015-04-15
Add all sub ref models for inline field

m0sth8:
If embeded(inlined) model has complex fields (like array of struct) then model builder should add all sub references.

### DIFF
--- a/swagger/model_builder.go
+++ b/swagger/model_builder.go
@@ -197,9 +197,11 @@ func (b modelBuilder) buildStructTypeProperty(field reflect.StructField, jsonNam
 			if required {
 				model.Required = append(model.Required, k)
 			}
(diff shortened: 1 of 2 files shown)


emicklei:
thank you for contributing. I will study this change asap

m0sth8:
How could I help you with tests? 
I can add only the sub models that don't existed yet.
The problem for me is that swagger validator can't work properly without sub models =(

emicklei:
Yes that would probably work. I am working on such a test now. The post build hook I mentioned earlier doesn't get called for submodels which can be considered a flaw on its own.

emicklei:
I will work on the flaw on top of your contribution instead.

emicklei:
And probably implement the "if not exists" approach

PR #211, opened 2015-06-13
Partial Content should pass

Nainterceptor:
I think this status must pass WriteHeader


Nainterceptor:
(The goal is to implement a range request from our API)

emicklei:
thank you for contributing

Issue #212, opened 2015-06-24
[RFC] Set Model.Description via a tag in a field of a struct


caesarxuchao:
go-restful doesn't yet populate the Model.Description [field](https://github.com/emicklei/go-restful/blob/master/swagger/swagger.go#L170). How about we make a new tag (e.g. "model-description") to allow users to specify this field? For example:

```
type computer struct {
    int cpu `json:"cpu" description:"cpu freq" model_description:"specification of a computer"`
    int memory
    ...
}
```
then in the `addModel()` function, we set the model's description to "specification of a computer".

We can discuss details like whether we should allow user to specify  "model_description" tag in multiple fields later.

I can code this up if people think this proposal is valid.

@emicklei @bgrant0607 @smarterclayton @nikhiljindal

bgrant0607:
We could just concatenate descriptions in the case of multiple.

Go naming conventions usually use camelCase rather than underscore.

FWIW, we plan to use this on embedded struct fields (TypeMeta).

Otherwise, SGTM

emicklei:
see https://github.com/emicklei/go-restful/commit/f949206d3742e97d3f4344e2db0ef97ebc28bafc for the result of merging and modifying the contribution of @gdamore

caesarxuchao:
@emicklei, thanks for the info. No, it's different. We want to set the Decription of a Model, not that of the ModelProperty. I will submit a patch.

emicklei:
Ok, well you could use the ModelBuildable interface for that. However, I won't work for embedded structs at the moment.  I am ok with the `modelDescription` tag.

If only Go had TypeTags...
```
type Person struct {
	Name string
} `description:"i represent anybody" `
```

caesarxuchao:
@emicklei, if I'm going to use ModelBuidlable interface, then I need to implement PostBuildModel() function for each struct that I want to add a modelDescription, don't? Plus, as you said, it wont' work for embedded structs, which is where we intend to add the modelDescription.

emicklei:
yes that's right. I think it's better to have/use the modelDescription tag instead.

caesarxuchao:
Thank you @emicklei.

emicklei:
See https://github.com/emicklei/go-restful/pull/215

PR #213, opened 2015-06-26
add modelDesription tag to populate Model.Description

caesarxuchao:
Add another field tag "modelDescription", which can be specified in 
In addModel(), contents in modelDescription tags are aggregated and populate to Model.Description

#212

@emicklei @bgrant0607


caesarxuchao:
@emicklei , just see your comment about ModelBuildable interface in #212. I will update the PR.

PR #215, opened 2015-07-08
Models can generate documentation through a struct method

andronat:
In [GoogleCloudPlatform/kubernetes](https://github.com/GoogleCloudPlatform/kubernetes) we are using Swagger to describe all of our types.

The problem that we are facing with the current way of describing our models is that we have to keep field and struct documentation in two different places (1) as comments 2) as struct tags so go-restful can parse). You can see [here](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/api/v1/types.go#L82) an example.

We think that it is easier for us to attach a method on each struct so go-restful can get the description from those methods instead of struct tags. We have a way to automatically parse the documentation from the source code and then create methods for each of our structs to contain the comments.

### DIFF
--- a/swagger/README.md
+++ b/swagger/README.md
@@ -21,9 +21,56 @@ Now, you can install the Swagger WebService for serving the Swagger specificatio
 	swagger.InstallSwaggerService(config)		
 	
 	
+Documenting Structs
+--
+
+Currently there are 2 ways to document your structs in the go-restful Swagger.
+
+###### By using struct tags
+- Use tag "description" to annotate a struct field with a description to show in the UI
+- Use tag "modelDescription" to annotate the struct itself with a description to show in the UI. The tag can be added in an field of the struct and in case that there are multiple definition, they will be appended with an empty line.
+
+###### By using the SwaggerDoc method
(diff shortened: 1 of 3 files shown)


andronat:
Hello @emicklei. To be honest I didn't see the `ModelBuildable` interface. But it is also true that we would like to not have a dependency. You think that this PR is not a good way to go?

emicklei:
@andronat , so I thought :-). 
I am happy with the contribution and also surpised to the updated Readme. Nice!

About the implementation, I noticed some reflection magic in there. Could that be avoided introducing a SwaggerDoc interface instead? See also https://github.com/emicklei/go-restful/blob/master/swagger/model_builder.go#L24

andronat:
I've made some changes to get rid of most of reflection. Unfortunately I couldn't find a way to keep the receiver of the function as a pointer :/

I also updated the documentation for spacial caveats.

emicklei:
@andronat,  thank you for updating the PR.  

Because SwaggerDoc is a read-only function, the receiver does not have to be pointer, so I don't think it is unfortunate.

PR #217, opened 2015-07-10
Paniced responses not sent with Container.EnableContentEncoding(true)

nullbus:
Since the `CompressingResponseWriter` is closed before calling panic handler, every response that panic handler make will not be sent to client. This PR fixes the situation.

Also I've added a test for this situation. The test will fail with previous code.

Thank you!


emicklei:
@nullbus , thank you for reporting and fixing this.

PR #226, opened 2015-08-22
Add the ability to remove a WebService to Container

brendandburns:


emicklei:
Thank you for this contribution. 

Interesting code fragment where you create n closure to encapsulate setting and releasing the lock around finding the best route. Never seen or used that one before. Nice.

Issue #228, opened 2015-09-05
Array return type not rendered properly


jonathathan:
Hi, I notice that running `examples/restful-user-service.go` from this repo produces a spec that swagger-ui doesn't seem to like. The resulting swagger-ui page is shown here:

![screen shot 2015-09-04 at 4 56 45 pm](https://cloud.githubusercontent.com/assets/7103322/9696546/2bc2bbd2-5326-11e5-9dc7-9e0a5afc650a.png)

The relevant source code, from `examples/restful-user-service.go`:

```    
ws.Route(ws.GET("/").To(u.findAllUsers).
        // docs
        Doc("get all users").
        Operation("findAllUsers").
        Returns(200, "OK", []User{}))
```

appears to generate this json:

```
  "apis": [
   {
    "path": "/users",
    "operations": [
     {
      "type": "void",
      "method": "GET",
      "responseMessages": [
       {
        "code": 200,
        "message": "OK",
        "responseModel": "array[main.User]"
       }
      ],
```

I played around with it a little and wrote a patch which produces the following, which I _believe_ to be proper Swagger 1.2 syntax (with User -> string for simplicity):

```
  "apis": [
   {
    "path": "/users",
    "operations": [
     {
      "type": "array",
      "items": {
        "type": "string"
      },
      "method": "GET",
      "responseMessages": [
       {
        "code": 200,
        "message": "OK"
       }
      ],
```

and swagger-ui didn't render any response schema, so there may be multiple facets to this bug.

Hope this helps! Also happy to lend a hand if I could be any help :)

emicklei:
thank you for reporting this. I will investigate the issue more.

emicklei:
could be related to https://github.com/emicklei/go-restful/issues/223 which is not solved by the way.

emicklei:
The role of type
https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#523-operation-object

emicklei:
At https://github.com/emicklei/go-restful/blob/master/swagger/swagger_webservice.go#L216 I see that it only takes into account the write sample which is specified by RouteBuilder.Write(interface{}). Now we need to decide what to do when both Returns and Writes have been called.

emicklei:
By searching for related Swagger problems I am starting to think that this is a unsupported feature in spec 1.2. Therefore I am strongly considering not fixing this and instead put my effort in getting support for 2.0 working.

jonathathan:
[...]

PR #234, opened 2015-09-23
fix bug with arrays in Writes() not parsed by swagger #228

vikstrous:
This allows you to specify an array in Writes(). (related to #228)

It looks like the spec doesn't support arrays in "responses" because they have to use a "responseModel" which is a string reference to a model. A model can't be an array AFAIK.

This is a good compromise because in most cases Writes() is sufficient for expressing the return type of an API.

I might not have written this patch in the best way possible, so feel free to completely rewrite it. I also didn't see an easy way to test the output, but at least I made sure that the code gets executed in a test case.

### DIFF
--- a/swagger/swagger.go
+++ b/swagger/swagger.go
@@ -134,7 +134,7 @@ type Api struct {
 
 // 5.2.3 Operation Object
 type Operation struct {
-	Type             string            `json:"type"`
+	DataTypeFields
 	Method           string            `json:"method"`
 	Summary          string            `json:"summary,omitempty"`
 	Notes            string            `json:"notes,omitempty"`
--- a/swagger/swagger_test.go
+++ b/swagger/swagger_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/emicklei/go-restful"
+	"github.com/emicklei/go-restful/swagger/test_package"
 )
 
 func TestInfoStruct_Issue231(t *testing.T) {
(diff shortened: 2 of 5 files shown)


vikstrous:
I found a bug with this PR. When using Writes() on a struct it doesn't get the name correctly. I'm not sure if I introduced it or this was always the case because I didn't change `reflect.TypeOf(any).Name()` in `asDataType()`.

vikstrous:
I fixed the bug. I also found out exactly why my fix works:

https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#523-operation-object
> This object includes the Data Type Fields in order to describe the return value of the operation. The type field MUST be used to link to other models.

vikstrous:
The test cases I have right now are actually wrong. They show what is currently produced, not what should be produced. It might require some refactoring to make this work for all cases.

vikstrous:
After testing things out in swagger-ui and reading the spec, it seems like in theory primitives should be supported here, but they are not in practice. Only arrays and objects are supported and apparently go-restful doesn't support maps even in structs, so I'll remove the tests for unsupported features and squash. Please re view my PR.

While I was working on this I also noticed that ReadSample is assigned, but not used. A "body" parameter is used instead.

emicklei:
Thanks, I will have a look. I am currently working on Issue #235 which is more urgent.

vikstrous:
PTAL... In particular, I'm interested in a more elegant way to achieve this: https://github.com/emicklei/go-restful/pull/234/files#diff-a76374c484db42c1ef8bd1d0fb232c9eR396

PR #238, opened 2015-09-25
Generate json #237

### LINKED ISSUE #237: Provide a way to generate swagger json programatically
Right now it seems like the only way to get the json output from the swagger generator is by hitting the web api. This is not good if you need to generate documentation using a command line tool. One can work around it by hitting the web api and storing the results, but there should be an easier way.

vikstrous:
fixes #237

based on / includes #234. I wanted to avoid merge conflicts.

### DIFF
--- a/swagger/swagger.go
+++ b/swagger/swagger.go
@@ -134,7 +134,7 @@ type Api struct {
 
 // 5.2.3 Operation Object
 type Operation struct {
-	Type             string            `json:"type"`
+	DataTypeFields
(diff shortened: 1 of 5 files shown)


emicklei:
thank you for contributing. As this is bigger change, I want to review and test this locally first ; probably later today (no promises)

vikstrous:
closing in favour of https://github.com/emicklei/go-restful/pull/242

Issue #239, opened 2015-09-29
Filters added after calling HandleWithFilter are not used


liggitt:
Since the receiver isn't a pointer, a snapshot of the filter slice is captured for use in the closure, and later additions to the container's filter chain aren't used

PR #240, opened 2015-09-29
Fix HandleWithFilter receiver

liggitt:
Fixes https://github.com/emicklei/go-restful/issues/239


emicklei:
Thank you for contributing and including a test for it.

PR #242, opened 2015-10-07
Generate json

### LINKED ISSUE #237: Provide a way to generate swagger json programatically

vikstrous:
fixes #237, based on / includes #234 to use the json array diff fix

I created a new SwaggerBuilder that can evolve independently from the SwaggerService. We can eventually change the implementation such that SwaggerService uses SwaggerBuilder rather than the other way around, but this works for now.

### DIFF
--- a/swagger/swagger.go
+++ b/swagger/swagger.go
@@ -134,7 +134,7 @@ type Api struct {
 
 // 5.2.3 Operation Object
 type Operation struct {
-	Type             string            `json:"type"`
+	DataTypeFields
 	Method           string            `json:"method"`
 	Summary          string            `json:"summary,omitempty"`
 	Notes            string            `json:"notes,omitempty"`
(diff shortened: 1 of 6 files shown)


emicklei:
thank you for your contribution!

emicklei:
"SwaggerService uses SwaggerBuilder " , I like that direction.  This will be the design for swagger 2.0 support in any case.

emicklei:
looking back at what has been merged, I noticed that some of my comments were not processed yet.  I may change them myself later ( e.g removing the xtra test package, commenting the builder ...)

vikstrous:
I don't think it's a good idea to remove the extra package. I was not able to test the package name prefix of the type without it. I had a bug related to that the first time I attempted to make Writes() work with arrays. And, yeah, sorry I forgot to comment the builder.

PR #243, opened 2015-10-07
Adds format to Parameter in go-restful for use in Swagger UI

gerbenjacobs:
Adds format to Parameter in go-restful and use it for the Format in Swagger UI.

This will fix issues such as the need to have dataType: "string" and dataFormat: "date-time"
https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#431-primitives

I made it "backwards compatible" as I'm not sure what the original plan was for the //TODO in asFormat()

Change-Id: I29bed78705e42a4c09c0bb85c08f0d232b1b40d0


emicklei:
thanks

PR #244, opened 2015-10-09
Removed left fmt.Println method in SwaggerService

mtojek:
Hi,

there was a smuggled fmt.Println method call in one of previous pull-requests, probably used for debugging. I removed this call.


emicklei:
thx for spotting and fixing this.

PR #250, opened 2015-11-02
Fixes #249 Path params with url-escaped forward slashes break routing.

### LINKED ISSUE #249: Path params with url-escaped forward slashes break routing.
Here is a snippet from url.go:224

// Note that the Path field is stored in decoded form: /%47%6f%2f becomes /Go/.
// A consequence is that it is impossible to tell which slashes in the Path were
// slashes in the raw URL and which were %2f. This distinction is rarely important,
// but when it is a client must use other routines to parse the raw URL or construct
// the parsed URL. For example, an HTTP server can consult req.RequestURI, and
// an HTTP client can use URL{Host: "example.com", Opaque: "//example.com/Go%2f"}
// instead of URL{Host: "example.com", Path: "/Go/"}.

So when I try to pass a URL-escaped string containing forward slashes as a path parameter, handler routing breaks and I see unpredictable behavior.  In my current configuration this results in an expected 404 Not Found.

jrosssavant:


emicklei:
can you please provide a test for this? thanks.

jrosssavant:
Looks like this change actually breaks a large number of existing tests because many rely on using http.NewRequest() which doesn't set the RequestURI field.  In fact the documentation in request.go suggests it is an error for the client to set RequestURI.

So I can't really suggest merging this change as-is given that you would like to keep your tests passing.  But re-writing a bunch of your tests to accommodate this change is not something I have time to do in the near term (nor am I sure you would accept such a change).  Can you suggest another solution to the problem described in #249?

emicklei:
I see, well in that case I am willing to correct the tests first w.r.t setting the RequestURI field ; don't want to use the http package in an unsupported way.  

Thank you for notifying this.

emicklei:
We still need a test first that demonstrates the problem. I will compose such as test

emicklei:
I wrote a litte program to play with parameters and route definitions.

```
package main

import (
	"io"
	"net/http"
	. "github.com/emicklei/go-restful"
)

func main() {
	DefaultContainer.Router(CurlyRouter{})
	ws := new(WebService)
	ws.Route(ws.GET("/slashy/{slash:*}").To(slash))
	Add(ws)
	http.ListenAndServe(":8080", nil)
}

func slash(req *Request, resp *Response) {
	io.WriteString(resp, "slash="+req.PathParameter("slash"))
}
```

`GET http://localhost:8080/slashy/Go%2Flang` ->  `400 Bad Request`
`GET http://localhost:8080/slashy/Go/lang`  ->  `slash=Go/lang`

emicklei:
feel free to reopen this if a (better) solution is found

usernameisnull:
> I wrote a litte program to play with parameters and route definitions.
> 
> ```
> package main
> 
> import (
>     "io"
>     "net/http"
>     . "github.com/emicklei/go-restful"
> )
> 
> func main() {
>     DefaultContainer.Router(CurlyRouter{})
>     ws := new(WebService)
>     ws.Route(ws.GET("/slashy/{slash:*}").To(slash))
>     Add(ws)
>     http.ListenAndServe(":8080", nil)
> }
> 
> func slash(req *Request, resp *Response) {
>     io.WriteString(resp, "slash="+req.PathParameter("slash"))
> }
> ```
> 
> `GET http://localhost:8080/slashy/Go%2Flang` -> `400 Bad Request` `GET http://localhost:8080/slashy/Go/lang` -> `slash=Go/lang`

{slash:*} ​​has to be at the end, right? If there are multiple such parameters in this url, it won't work, right?

emicklei:
running your program
```
$ curl http://localhost:8080/slashy/Go%2Flang
slash=Go/lang
```

PR #252, opened 2015-11-13
Proposed addition of 'type' tag for swagger

nathanejohnson:
Currently I have some struct fields that are essentially a type alias for an integer.  They do not show up in my swagger documentation when I pass them to Reads(), they are omitted entirely.  In the long term, it might be nice for the reflection magic in the swagger model_builder etc could detect these and handle them as though they were the appropriate type.  Implementing a MarshalJSON method on my aliased type did get them to show up in the swagger documentation, but always as type "string" versus integer.  This PR is much less ambitious than a fix for the above, but could prove generically useful in the event that we ever wanted to override the type that was guessed.

### DIFF
--- a/swagger/model_property_ext.go
+++ b/swagger/model_property_ext.go
@@ -31,6 +31,12 @@ func (prop *ModelProperty) setMaximum(field reflect.StructField) {
 	}
 }
 
+func (prop *ModelProperty) setType(field reflect.StructField) {
+	if tag := field.Tag.Get("type"); tag != "" {
+		prop.Type = &tag
+	}
+}
+
 func (prop *ModelProperty) setMinimum(field reflect.StructField) {
 	if tag := field.Tag.Get("minimum"); tag != "" {
 		prop.Minimum = tag
@@ -56,4 +62,5 @@ func (prop *ModelProperty) setPropertyMetadata(field reflect.StructField) {
 	prop.setMaximum(field)
 	prop.setUniqueItems(field)
 	prop.setDefaultValue(field)
+	prop.setType(field)
 }
--- a/swagger/model_property_ext_test.go
+++ b/swagger/model_property_ext_test.go
@@ -4,11 +4,13 @@ import "testing"
 
 // clear && go test -v -test.run TestThatExtraTagsAreReadIntoModel ...swagger
 func TestThatExtraTagsAreReadIntoModel(t *testing.T) {
+	type fakeint int
 	type Anything struct {
-		Name     string `description:"name" modelDescription:"a test"`
-		Size     int    `minimum:"0" maximum:"10"`
-		Stati    string `enum:"off|on" default:"on" modelDescription:"more description"`
-		ID       string `unique:"true"`
+		Name     string  `description:"name" modelDescription:"a test"`
+		Size     int     `minimum:"0" maximum:"10"`
+		Stati    string  `enum:"off|on" default:"on" modelDescription:"more description"`
+		ID       string  `unique:"true"`
+		FakeInt  fakeint `type:"integer"`
 		Password string
 	}
 	m := modelsFromStruct(Anything{})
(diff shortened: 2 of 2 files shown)


nathanejohnson:
Closing this for now, this doesn't work if the type is already guessed.  Going to try to see if there is a better approach

emicklei:
Thank you for adding this feature ; I can see that developers now have more control over the documented type.

emicklei:
ok ic

oschmid:
Hi, I'm also running into the issue of "aliased" types being hidden in swagger. Can you recommend a way to get types to show up in swagger documentation as their underlying type (e.g. time.Duration -> integer)? Thanks in advance @nathanejohnson @emicklei .

emicklei:
We could add this type to the mapping as defined in `jsonSchemaType` (model_builder.go)

nathanejohnson:
oschmid, just now saw this.  Anyway, doesn't `type:"integer"` annotation work ?

from model_builder_test.go:

// This tests a primitive with type overrides in the struct tags
type FakeInt int
type E struct {
        Id FakeInt `type:"integer"`
        IP net.IP  `type:"string"`
}

There was another pull request that was accepted separately from this one that implemented this functionality mentioned in this particular PR, but the functionaly should be in master.

oschmid:
@nathanejohnson Annotating with `type:"integer"` works. Thanks!

oschmid:
@nathanejohnson I'm running into a similar issue with arrays of FakeInt's. I've tried the following type annotation without success.
````
type FakeInt int
type E struct {
    Id []FakeInt `type:"Array[string]"`
}
````
It generates a correct swagger UI "Model":
````
E {
    Id (Array[string])
}
````
But not a correct swagger UI "Example Value":
````
{
}
````
Is there an alternate type annotation format I should be using? (Also apologies if this is the wrong place to raise this issue but I could't find a more related one)

**Comment by nathanejohnson** [comment]
I'm not entirely sure what you're wanting, but i believe you might be able to get what you want with the "default" annotation.

**Comment by oschmid** [comment]
Sorry @nathanejohnson , I made some typos in my comment. What I'm looking for is a way for an array of FakeString's look like an array of string's in swagger.
Without any annotation:
````
type FakeString string
type E struct {
    Id []FakeString
}
````
The type of E.Id shown in the swagger UI Model documentation is "Array[FakeString]":
````
E {
    Id (Array[FakeString])
}
````
And an empty struct is shown in the swagger UI Example Value:
````
{
    Id [
        {}
    ]
}
````

With the following type annotation, the model looks like the model of a regular string array field but the example value is missing the field completely.
````
type FakeString string
type E struct {
    Id []FakeString `type:"Array[string]"`
}
````
Generates a seemingly correct swagger UI "Model":
````
E {
    Id (Array[string])
}
````
But an incorrect swagger UI "Example Value":
````
{
}
````
What I'd expect with this annotation is the following swagger UI "Example Value":
````
{
    Id: [
        "string"
    ]
}
````
Is that more helpful? If you could point me in the direction of where the type annotations are processed I could take a look at it myself as well.

**Comment by nathanejohnson** [comment]
@oschmid Yes this is more clear.  does the "default" annotation not give you what you want there?

**Comment by oschmid** [comment]
What do you mean by "default" annotation? I've tried no annotation and `type:"default"` and both fail to generate an array of strings in the example value.

**Comment by nathanejohnson** [comment]
default is an annotation separate from type.  so you'd try something like 


```
type FakeString string
type E struct {
    Id []FakeString `type:"Array[string]" default:"[\"string\"]"`
}
```

oschmid:
That's closer. It gives the following example value:
````
{
    Id: "[\"string\"]"
}
````
It's not exactly the same (the backslashes are not escaped, "string" is not on a new line, and the whole thing is in quotes). `type:"Array[string]" default:"[string]"` gives slightly clearer results:
````
{
    Id: "[string]"
}
````
Still not perfect but good enough for my purposes. Thanks for you help!

PR #253, opened 2015-11-13
Adding 'type' struct tag that can override the guessed type

nathanejohnson:
This is a more complete version of my earlier PR.  This adds a "type" struct tag that can override the type guessed by reflection.  I have updated the unit tests with some examples of where this might be useful, as an example a net.IP type and an int type alias.  net.IP is itself a type alias for []byte, but it implements json.Unmarshaler and json.Marshaler to allow it to be represented as a string or converted from a string.

### DIFF
--- a/swagger/model_builder.go
+++ b/swagger/model_builder.go
@@ -132,9 +132,11 @@ func (b modelBuilder) buildProperty(field reflect.StructField, model *Model, mod
 		modelDescription = tag
 	}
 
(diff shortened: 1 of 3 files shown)


nathanejohnson:
Still more work to do apparently.

nathanejohnson:
Found a typo in the original PR that caused a bug.  tests are passing once again.

PR #254, opened 2015-11-13
Adding 'type' struct tag that can override the guessed type

nathanejohnson:
This adds a "type" struct tag that can override the type guessed by reflection. I have updated the unit tests with some examples of where this might be useful, as an example a net.IP type and an int type alias. net.IP is itself a type alias for []byte, but it implements encoding.TextUnmarshaler and encoding.TextMarshaler to allow it to be represented as a string or converted from a string.

By the way, sorry for the closed then new PR's.  I didn't realize that if you close a PR, then rebase and force push, it wouldn't let you reopen.

### DIFF
--- a/swagger/model_builder.go
+++ b/swagger/model_builder.go
@@ -132,9 +132,11 @@ func (b modelBuilder) buildProperty(field reflect.StructField, model *Model, mod
 		modelDescription = tag
 	}
 
-	fieldType := field.Type
-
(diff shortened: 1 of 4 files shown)


nathanejohnson:
I just added one more unit test in model_builder_test.go to cover this type override tag.

emicklei:
Thank you for contributing to this package.  I reviewed the changes and tests; LGTM.

PR #255, opened 2015-11-17
Put acquired resources back to compressor pool

mtojek:
Hi Ernest,

I was curious about why do not put back resources to the compressor pool? Please review changes and have a nice day.

Best regards,
Marcin


emicklei:
Thank you for pointing this out.

The documentation of sync.Pool is not clear about the use of Put. In could be used to fill the pool but, as I understand it, the pool will only keep those resources in between GC calls (but not garanteed).

```Any item stored in the Pool may be removed automatically at any time without notification. If the Pool holds the only reference when this happens, the item might be deallocated.```

I think it is safe to put it back to the pool so I will probably accept your change.  (why was first commit included?)

Issue #236, opened 2015-09-25
Access to the Route from within a Request


vikstrous:
The only information a request can get about its route seems to be SelectedRoutePath(). It would be very helpful to also have a Route() function on a Request object to get the actual Route object. I want this so that I can log documentation violations at run time. This feature can also enable writing a check for making sure all documented responses have been tested.

emicklei:
The request currently has the selectedRoutePath ; we could replace that with a reference to the Route object and refactor the code unsing this Path without growing the Request struct.

However, how can we prevent applications from modifying this Route ; not all changes are reflected in the underlying datastructure (Router, SwaggerService). How do see that?

vikstrous:
I don't think we need to prevent the application from modifying the route. Just because they can doesn't mean they will... you can just mention in the documentation that they shouldn't.

emicklei:
@vikstrous can we close this?

PR #257, opened 2015-11-23
experiment in adding optional magic

vikstrous:
Do not merge this as it is. This is an experiment I did in adding some optional magic. I want to see what people think about this. Is it useful? Is it too specific? Do the function signatures make sense?

I haven't written `typeIsRequestFunc` yet, but I think you can imagine what that would look like.

The basic idea is to use `.Magic()` instead of `.To()` and have the router handle the boilerplate for reading and writing responses. This does not replace `To` because it can't handle more advanced use cases, but I think it would be nice to have.

### DIFF
--- a/examples/restful-user-resource-magic.go
+++ b/examples/restful-user-resource-magic.go
@@ -1,0 +1,137 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"strconv"
+
+	"github.com/emicklei/go-restful"
+	"github.com/emicklei/go-restful/swagger"
+)
+
+// This example show a complete (GET,PUT,POST,DELETE) conventional example of
+// a REST Resource including documentation to be served by e.g. a Swagger UI
+// It is recommended to create a Resource struct (UserResource) that can encapsulate
+// an object that provide domain access (a DAO)
+// It has a Register method including the complete Route mapping to methods together
+// with all the appropriate documentation
+//
+// POST http://localhost:8080/users
+// <User><Id>1</Id><Name>Melissa Raspberry</Name></User>
+//
+// GET http://localhost:8080/users/1
+//
+// PUT http://localhost:8080/users/1
+// <User><Id>1</Id><Name>Melissa</Name></User>
+//
+// DELETE http://localhost:8080/users/1
+//
+
+type User struct {
+	Id, Name string
+}
+
+type UserResource struct {
+	// normally one would use DAO (data access object)
+	users map[string]User
+}
+
+func (u UserResource) Register(container *restful.Container) {
+	ws := new(restful.WebService)
+	ws.
+		Path("/users").
+		Doc("Manage Users").
+		Consumes(restful.MIME_XML, restful.MIME_JSON).
+		Produces(restful.MIME_JSON, restful.MIME_XML) // you can specify this per route as well
+
+	ws.Route(ws.GET("/{user-id}").Magic(u.findUser).
+		// docs
+		Doc("get a user").
+		Operation("findUser").
+		Param(ws.PathParameter("user-id", "identifier of the user").DataType("string")).
+		Writes(User{})) // on the response
+
+	ws.Route(ws.PUT("/{user-id}").Magic(u.updateUser).
+		// docs
+		Doc("update a user").
+		Operation("updateUser").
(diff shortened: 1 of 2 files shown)


emicklei:
I will have a look at the code and think about this idea. Thanks

nathanejohnson:
This is kind of interesting.  So is the main point of this to take the pain away from unmarshaling data into generic structs and handle errors consistently?  If so, I ended up doing something that handles unmarshaling request bodies with a filter.  Here is an example

    // This function returns a go-restful filter function that unmarshals data from a POST or PUT body
    // into a value of type DataType, then stores the resulting value in a request attribute
    // with name "name".  This allocates a new variable the same type as DataType on your behalf.
    // DataType is never read or modified outside of inspecting its type, so it is not a requirement
    // that an allocated variable be passed in.  Because of this, passing (*Foo)(nil) is valid,
    // and the attribute will point to an allocated instance of Foo populated with input.
    //
    // Valid DataType types are are structs and pointers to structs.  If you pass a struct, the attribute
    // will be a struct, if you pass a pointer, the attribute will be a pointer to the struct.  If a bad
    // DataType is passed in, a runtime panic will be generated.  In case of an error when calling
    // ReadEntity, returns a bad request error and stops the chain.
    //
    // Example:
    //
    //		ws.Route(ws.POST("/foo").
    //			To(inst.CreateFoo).
    //			Reads(Foo{}).
    //			Doc("Create a foo record").
    //			Filter(this.UnmarshalBodyFilter((*Foo)(nil), "foo")).
    //			Writes(ResponseStatus{}))
    //
    //		func (inst *Instance) CreateFoo(req *restful.Request, resp *restful.Response) {
    //			foo := req.Attribute("foo").(*Foo)
    //			// if we get here, foo is guaranteed to be non-nil
    //			// do useful stuff
    //		}
    func (this *Instance) UnmarshalBodyFilter(DataType interface{}, name string) restful.FilterFunction {
    	return func(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
    		// A bunch of code to make sure we don't deref a nil pointer or other accidental panic,
    		// instead sending our own custom awesome panic.  Don't panic!
    		dtType := reflect.TypeOf(DataType)
    		if dtType == nil {
    			panic("nil passed as DataType to UnmarshalBodyFilter")
    		}
    
    		dtOrigType := dtType
    		wantsPtr := false
    
    		// Check to see if this DataType is a pointer, if so deref once
    		if dtType.Kind() == reflect.Ptr {
    			dtType = dtType.Elem()
    			wantsPtr = true
    		}
    
    		// Ensure this is a struct
    		if dtType.Kind() != reflect.Struct {
    			panic(fmt.Sprintf("Invalid DataType passed to UnmarshalBodyFilter: %s",
    				dtOrigType.Kind().String()))
    		}
    
    		// Now we get down to business of being marginally useful.
    		// Allocate a new struct
    		data := reflect.New(dtType)
    
    		// Unmarshal / decode
    		err := req.ReadEntity(data.Interface())
    
    		// All this work just for consistent error handling...
    		if err != nil {
    			errorBadRequest.WriteStatus(resp, err.Error())
    			return
    		}
    
    		// Success!
    		// data represents a pointer value, so figure out what to do
    		if wantsPtr {
    			req.SetAttribute(name, data.Interface())
    		} else {
    			req.SetAttribute(name, data.Elem().Interface())
    		}
    		this.Log.Debug("received data: \n%#v", req.Attribute(name))
    
    		chain.ProcessFilter(req, resp)
    	}
    }

vikstrous:
Considering that the whole point of this framework is to not have magic, I like your filter approach. I would go a little further and say that it should be possible to know the type of the request from the `Reads()` documentation and the name of the body struct should always be the same. Then this filter can be applied globally with no arguments.

Writing this magic unmarshalling filter would require access to the Route in the filter https://github.com/emicklei/go-restful/issues/236

Maybe this filter can be provided by go-restful and be trivial to optionally include your app in one line.

vikstrous:
One thing that's awkward with both of our approaches is that the error to return in case of input parsing failure can't be customized. We might need better error reporting infrastructure to go along with this input parser. Maybe something like what I proposed above might work. I don't have a specific implementation in mind yet. I'll open a separate issue for that.

emicklei:
i guess this work is no longer of interest to you, close it?

PR #258, opened 2015-11-25
Fix: nil pointer dereference when writing using closed compressor

mtojek:
Hi,

having a defined container with enabled compression, it may occur an accidential situation when a Write operation is perfomed after a Close (compressor) operation. The result of such operation order (Write - Close - Write) is a nil ptr dereference and application "tango down" ;)
I've added the compressor presence check.

Best regards,
Marcin


emicklei:
Thank you for reporting your issue and safety fix.

PR #262, opened 2015-12-17
Add Flush method to response

ddysher:
Come across this when tried to flush buffered data to client.  If this is not a wanted feature, do we have a better way to achieve the goal?


emicklei:
Thank you for this contribution.

ddysher:
:+1: Great tool!

PR #267, opened 2016-02-02
Implement http.Hijacker interface on CompressingResponseWriter

schleppy:
When using a single `Container` that included routes for websockets it was discovered that setting `EnableContentEncoding(true)` was causing an error with gorilla's `websocket.Upgrader` as the `CompressingResponseWriter` does not implement the `http.Hijacker` interface.  This change adds a simple Hijack implementation that allows for the use of websocket endpoints within a `Container` when `EnableContentEncoding(true)` is set.

### DIFF
--- a/compress.go
+++ b/compress.go
@@ -9,7 +9,6 @@ import (
 	"compress/gzip"
 	"compress/zlib"
 	"errors"
-	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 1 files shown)


emicklei:
Thank you for contributing. First time I hear about using it with websockets. nice

PR #268, opened 2016-02-08
fix some example import paths

Manbeardo:
Some examples referenced google.golang.com instead of google.golang.org


emicklei:
thanks

PR #271, opened 2016-03-22
Fix invalid compared Swagger paths

mtojek:
Hi,

I have defined two webservices:

/foobar
/foo

Each of them has defined GET and HEAD routes. When listing swagger API declarations, endpoints corresponding to /foo are also listed under /foobar.

Fix checks if there are compared entire paths instead of sudden exact substrings.

Please review the pull request.

Kind regards,
Marcin

### DIFF
--- a/swagger/swagger_webservice.go
+++ b/swagger/swagger_webservice.go
@@ -226,10 +226,9 @@ func (sws SwaggerService) composeDeclaration(ws *restful.WebService, pathPrefix 
 	pathToRoutes := newOrderedRouteMap()
 	for _, other := range ws.Routes() {
 		if strings.HasPrefix(other.Path, pathPrefix) {
-			if len(other.Path) > len(pathPrefix) && other.Path[len(pathPrefix)] != '/' {
+			if len(pathPrefix) > 1 && len(other.Path) > len(pathPrefix) && other.Path[len(pathPrefix)] != '/' {
 				continue
 			}
-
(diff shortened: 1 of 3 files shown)


emicklei:
thank you for contributing. please include a test that demonstrates the fix

mtojek:
I've just attached a test case.

@rafos keep watching!

swagger/swagger_webservice.go
```diff
@@ -226,10 +226,9 @@ func (sws SwaggerService) composeDeclaration(ws *restful.WebService, pathPrefix
 	pathToRoutes := newOrderedRouteMap()
 	for _, other := range ws.Routes() {
 		if strings.HasPrefix(other.Path, pathPrefix) {
-			if len(other.Path) > len(pathPrefix) && other.Path[len(pathPrefix)] != '/' {
+			if len(pathPrefix) > 1 && len(other.Path) > len(pathPrefix) && other.Path[len(pathPrefix)] != '/' {
```
emicklei: the condition is hard to read and therefore difficult to verify whether all cases are caught.
for this reason, it might be better to define a simple function for it that takes a 2 paths and a prefix and return a bool. 
Then it will be easy to write tests for this and see, by inspecting the samples, that the original issue is solved.

emicklei:
thank you for contributing

mtojek:
That was a long time ago, anyway thanks!

Issue #278, opened 2016-04-13
Content-Type required on empty POST


ReneKroon:
RFC 7231 does not state that an empty payload should have a Content-Type header:

   A sender that generates a message containing a payload body SHOULD
   generate a Content-Type header field in that message unless the
   intended media type of the enclosed representation is unknown to the
   sender.  If a Content-Type header field is not present, the recipient
   MAY either assume a media type of "application/octet-stream"
   ([RFC2046], Section 4.5.1) or examine the data to determine its type.

Currently i get a http 415 on these POST requests without a body. JSR-311 does allow this though. Im not sure how to fix this at this time as any Consumes(restful.MIME*) route would be valid for these kinds of requests.

emicklei:
you can set it like this 
```
.Consumes("application/octet-stream")
```

emicklei:
please have a look at the test and if not convinced, do provide a test that demonstrates this issues.

ReneKroon:
was fixed withe .Consumes(.. part

PR #280, opened 2016-04-13
Clarify that a 406 might occur on a POST with empty body and how to fix

ReneKroon:
lets document this case, #278 i closed the other one.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -131,7 +131,9 @@ The request has a valid URL but the method (GET,PUT,POST,...) is not allowed.
 
 	406: Not Acceptable
 
-The request does not have or has an unknown Accept Header set for this operation.
+The request does not have or has an unknown Accept Header set for this operation. Note that a 406 might also occur
+when doing a POST request with empty body. The restful application should use .Consumes("application/octet-stream")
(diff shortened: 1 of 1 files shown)


doc.go
```diff
@@ -131,7 +131,9 @@ The request has a valid URL but the method (GET,PUT,POST,...) is not allowed.
 
 	406: Not Acceptable
 
-The request does not have or has an unknown Accept Header set for this operation.
+The request does not have or has an unknown Accept Header set for this operation. Note that a 406 might also occur
```
emicklei: please clarify the word "might" in this context.  IMHO it either works or not. Could you provide an example testcase that demonstrates the problem and solution?

ReneKroon: The issued described is that on the server side you need to specify a .Consumes() clause even on calls with an empty body. So it would be something like 'Note that a 406 might also occur if no .Consumes("application/octet-stream") statement has been set on the endpoint'. I'll check whether application/octet-stream is an example like json, xml or required configuration for empty calls.

emicklei: feel free to update/reopen it

PR #282, opened 2016-04-16
Move msgpack to example

bailantaotao:
i have moved the msg pack to example folder,
but i don't know the const MIME_MSGPACK = "application/x-msgpack" place in here is OK?

### DIFF
--- a/examples/msgpack/msgpack_entity.go
+++ b/examples/msgpack/msgpack_entity.go
@@ -9,14 +9,12 @@ const MIME_MSGPACK = "application/x-msgpack" // Accept or Content-Type used in C
 
 // NewEntityAccessorMPack returns a new EntityReaderWriter for accessing MessagePack content.
 // This package is not initialized with such an accessor using the MIME_MSGPACK contentType.
-func NewEntityAccessorMsgPack(contentType string) restful.EntityReaderWriter {
-	return entityMsgPackAccess{ContentType: contentType}
+func NewEntityAccessorMsgPack() restful.EntityReaderWriter {
+	return entityMsgPackAccess{}
 }
 
 // entityOctetAccess is a EntityReaderWriter for Octet encoding
 type entityMsgPackAccess struct {
-	// This is used for setting the Content-Type header when writing
-	ContentType string
 }
 
 // Read unmarshalls the value from byte slice and using msgpack to unmarshal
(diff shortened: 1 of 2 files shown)


examples/msgpack/msgpack_entity.go
```diff
@@ -0,0 +1,42 @@
...
+	restful "github.com/emicklei/go-restful"
+	"gopkg.in/vmihailenco/msgpack.v2"
+)
+
+const MIME_MSGPACK = "application/x-msgpack" // Accept or Content-Type used in Consumes() and/or Produces()
+
+// NewEntityAccessorMPack returns a new EntityReaderWriter for accessing MessagePack content.
+// This package is not initialized with such an accessor using the MIME_MSGPACK contentType.
+func NewEntityAccessorMsgPack(contentType string) restful.EntityReaderWriter {
+	return entityMsgPackAccess{ContentType: contentType}
```
emicklei: do you really need the ContentType to be configurable?  The XML version needed this field to specify character encoding. You probably use the fixed const.

examples/msgpack/msgpack_entity_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	err := resp.WriteEntity(mpack)
+	if err != nil {
+		t.Errorf("err %v", err)
+	}
+
+	// Read
+	bodyReader := bytes.NewReader(httpWriter.Body.Bytes())
+	httpRequest, _ := http.NewRequest("GET", "/test", bodyReader)
+	httpRequest.Header.Set("Content-Type", "application/x-msgpack; charset=UTF-8")
```
emicklei: does it make sense to pass the charset here?  msgpack is binary protocol.  If not then the accessor probably does not need the contentType field

bailantaotao: yes, you're right. i has removed charset.

examples/msgpack/msgpack_entity.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// entityOctetAccess is a EntityReaderWriter for Octet encoding
+type entityMsgPackAccess struct {
+	// This is used for setting the Content-Type header when writing
+	ContentType string
```
emicklei: do you still need the ContentType field?

bailantaotao: this is needed by unit test that can check the content-type.

i found a problem.
in my case, i have called response.writeAsJson to response client message,
but if i want response data with message pack format, i cannot do this, because response.go not contain any method about that...

i think i must folk your project then add this method then reference this folk that i can response client with message pack format.
```
func (r *Response) WriteAsMsgPack(value interface{}) error {
	return writeAsMsgPack(r, http.StatusOK, MIME_MSGPACK, value)
}
```
do you have any comments?

emicklei: I don't think you need this change.  The entity api of go-restful request handling consist of ReadEntity and WriteEntity. Those methods respect the Accept header and the Produces mime types. You should not have to call e.g. WriteAsMsgPack directly.

You test example shows that

```
resp.WriteEntity(mpack)
```

bailantaotao: Thanks you for reminding.

I have not call response.SetRequestAccepts(restPack.MIME_MSGPACK) when response.WriteEntity(interface{}) is doing....

so in my service, received a `Content-Type: application/json` always

i have remove the content-type! thanks!

emicklei: I think you get ContentType json because you are skipping a part of go-restful in your test.

Perhaps you should write a simple example program, similar to the hello-world, that create a WebService, sets both Consumes and Produces to msgpack and show how the handler works.

bailantaotao: hello, i have added the sample to show how to use.

emicklei:
Thank you @bailantaotao for adding this example use of go-restful.

PR #286, opened 2016-04-30
cors filter support regex check against allowd origin

AdoHe:
@emicklei ptal. I have a concern about the performance, I want to keep the regexp compile at most once. Hope to get your opinion.

### DIFF
--- a/cors_filter.go
+++ b/cors_filter.go
@@ -5,6 +5,7 @@ package restful
 // that can be found in the LICENSE file.
 
 import (
+	"regexp"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


cors_filter.go
```diff
@@ -19,11 +20,13 @@ import (
...
 	ExposeHeaders  []string // list of Header names
 	AllowedHeaders []string // list of Header names
-	AllowedDomains []string // list of allowed values for Http Origin. If empty all are allowed.
+	AllowedDomains []string // list of allowed values for Http Origin. An allowed value can be a regular expression to support subdomain matching. If empty all are allowed.
 	AllowedMethods []string
 	MaxAge         int // number of seconds before requiring new Options request
 	CookiesAllowed bool
 	Container      *Container
+
+	AllowedOriginPatterns []*regexp.Regexp // internal field for origin regex check.
```
emicklei: internal fields can be un-exposed

emicklei:
one optimization i can think of is to check first against raw string values of AllowedDomains. If there is a match then you are done. AllowedDomain may or may not use a regex notation. If there is no match then visit all the cached allowedPatterns with matchings.

AdoHe:
Just apply this optimization. @emicklei

cors_filter.go
```diff
@@ -37,15 +40,37 @@ func (c CrossOriginResourceSharing) Filter(req *Request, resp *Response, chain *
 		chain.ProcessFilter(req, resp)
 		return
 	}
-	if len(c.AllowedDomains) > 0 { // if provided then origin must be included
-		included := false
-		for _, each := range c.AllowedDomains {
-			if each == origin {
-				included = true
+	if len(c.AllowedDomains) > 0 { // if provided then origin must be allowed
```
emicklei: seeing some duplicate code here and in isOriginAllowed, is it possible to call that method here instead?

AdoHe:
@emicklei just reuse some code, ptal.

emicklei:
thanks

PR #287, opened 2016-05-02
Don't re-register dispatch handlers

brendandburns:


emicklei:
I think with the recent merge of branch Issue274, we no longer this PR

PR #290, opened 2016-05-08
fix compile path twice issue

AdoHe:
@emicklei ptal.
This will change `Path("") --> compilePathExpression() --> Path("/") --> compilePathExpression()` to `Path("") --> compilePathExpression()`

### DIFF
--- a/web_service.go
+++ b/web_service.go
@@ -36,9 +36,6 @@ func (w *WebService) SetDynamicRoutes(enable bool) {
 
 // compilePathExpression ensures that the path is compiled into a RegEx for those routers that need it.
 func (w *WebService) compilePathExpression() {
-	if len(w.rootPath) == 0 {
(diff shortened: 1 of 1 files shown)


web_service.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/emicklei/go-restful/log"
 )
 
+const DefaultRootPath = "/"
```
emicklei: i don't think introducing an exposed (part of api) const is needed. It will probably never be changed and I would argue it to be a recommended way to globally change routes of an app.

AdoHe: I did agree that we shouldn't expose a const here, so just remove it. This PR will not change the default behavior of `WebService`, but just fix an edge case when you pass an empty string to `Path`.

emicklei:
thanks @AdoHe

PR #291, opened 2016-06-02
wrong package breaks vendoring (gb)

DMXRoid:
The package for examples/restful-curly-router_test.go file was set to examples, instead of main like the rest of the files in that directory, which was breaking gb vendoring.


emicklei:
thanks

Issue #296, opened 2016-06-17
Specify swagger format field for custom marshaled model properties


mbohlool:
In kubernetes we are using a custom time field (unversioned.Time) to format the json output string for time. The problem is the generated Swagger specs will not set the format to 'date-time' for that property. We can fix that by providing another method (e.g. GetSwaggerFormat or even a more generalized method like SetSwaggerProperties) that returns the appropriate format. BuildProperty method in model_builder.go can call that method if it is supported for that property and the format field is missing.
An alternative way is to provide an static method to model builder that controls setting swagger specs for a property. Much like ModelProperty.setPropertyMetadata, but user provided (e.g. if the user provided method set type, we would skip the rest of the BuildProperty method.)

related issue: https://github.com/kubernetes/kubernetes/issues/2968

PR #297, opened 2016-06-22
Customizable typename-to-swagger-format-field map

mbohlool:
For a type that is marshaled manually, we do not have a way to specify its format. For example if we have a custom_time type that does not provided a custom marshaling, its swagger type would be string and its swagger format would be empty. The correct swagger format in this example is "date-time". This PR adds a function that maps type-name to swagger-format-field. If the function exists in swagger.Config and returns non-empty strings, that value will be used instead of internal model-builder map. #296

### DIFF
--- a/swagger/config.go
+++ b/swagger/config.go
@@ -9,6 +9,8 @@ import (
 // PostBuildDeclarationMapFunc can be used to modify the api declaration map.
 type PostBuildDeclarationMapFunc func(apiDeclarationMap *ApiDeclarationList)
 
+type MapSchemaFormatFunc func(typeName string) string
+
 type Config struct {
 	// url where the services are available, e.g. http://localhost:8080
 	// if left empty then the basePath of Swagger is taken from the actual request
@@ -31,4 +33,6 @@ type Config struct {
 	PostBuildHandler PostBuildDeclarationMapFunc
 	// Swagger global info struct
 	Info Info
+	// [optional] If set, model builder should call this handler to get addition typename-to-swagger-format-field convertion.
+	SchemaFormatHandler MapSchemaFormatFunc
 }
--- a/swagger/model_builder.go
+++ b/swagger/model_builder.go
@@ -14,6 +14,7 @@ type ModelBuildable interface {
 
 type modelBuilder struct {
 	Models *ModelList
+	Config *Config
 }
 
 type documentable interface {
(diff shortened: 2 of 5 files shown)


swagger/config.go
```diff
@@ -9,6 +9,8 @@ import (
 // PostBuildDeclarationMapFunc can be used to modify the api declaration map.
 type PostBuildDeclarationMapFunc func(apiDeclarationMap *ApiDeclarationList)
 
+type GetAdditionalSchemaFormat func(typeName string) string
```
emicklei: Suggestion to rename it as MapSchemaFormatFunc  ,  the other type also has the Func suffix and the acutal behavior is mapping (that could be implemented by a get on a hash)

swagger/config.go
```diff
@@ -31,4 +33,6 @@ type Config struct {
 	PostBuildHandler PostBuildDeclarationMapFunc
 	// Swagger global info struct
 	Info Info
+	// [optional] If set, model builder should call this handler to get addition typename-to-swagger-format-field convertion.
+	AdditionalFormatHandler GetAdditionalSchemaFormat
```
emicklei: I do not like the "Additional" prefix. Why not just SchemaFormatHandler?

swagger/config.go
mbohlool: Done.

swagger/config.go
mbohlool: You are right. the whole "additional" prefix was not a good idea. Fixed. Does it make sense to send  new clean PR?

emicklei: Its ok the change the current PR or squash it later

On Thu, Jun 23, 2016 at 10:49 AM, mbohlool <notifications@github.com> wrote:

> In swagger/config.go
> <https://github.com/emicklei/go-restful/pull/297#discussion_r68197246>:
>
> > @@ -31,4 +33,6 @@ type Config struct {
> >  	PostBuildHandler PostBuildDeclarationMapFunc
> >  	// Swagger global info struct
> >  	Info Info
> > +	// [optional] If set, model builder should call this handler to get addition typename-to-swagger-format-field convertion.
> > +	AdditionalFormatHandler GetAdditionalSchemaFormat
>
> You are right. the whole "additional" prefix was not a good idea. Fixed.
> Does it make sense to send new clean PR?
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/emicklei/go-restful/pull/297/files/122305779cf3fc57bc9e0c9a28a38951c362b65c#r68197246>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAp4jO5uKIPaVANzC8ZYqe33pVKx8LX6ks5qOki3gaJpZM4I8P1o>
> .
>



-- 
Met vriendelijke groet,

^ Ernest Micklei

ernestmicklei.com

mbohlool: I changed the last commit to fix this. Would you please take another look.

emicklei:
thank you for contributing @mbohlool

mbohlool:
Thanks for the merge. Look like I missed to pass swagger config in one place. Sorry for that. sent PR.

PR #303, opened 2016-07-03
Add "Info" to go-restful ApiDecl

smarterclayton:
Allows post processing of ApiDecl to set the info field for swagger 1.0
(to add things like descriptions of APIs).


emicklei:
thanks @smarterclayton

PR #311, opened 2016-08-08
swagger: report golang type []byte as json type string

mikedanese:
cc @liggitt @nikhiljindal

### DIFF
--- a/swagger/model_builder.go
+++ b/swagger/model_builder.go
@@ -51,6 +51,14 @@ func (b modelBuilder) addModel(st reflect.Type, nameOverride string) *Model {
 	if b.isPrimitiveType(modelName) {
 		return nil
 	}
+	// golang encoding/json packages says array and slice values encode as
+	// JSON arrays, except that []byte encodes as a base64-encoded string.
+	// If we see a []byte here, treat it at as a primitive type (string)
+	// and deal with it in buildArrayTypeProperty.
+	if (st.Kind() == reflect.Slice || st.Kind() == reflect.Array) &&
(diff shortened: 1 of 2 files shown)


swagger/model_builder.go
```diff
@@ -168,8 +172,14 @@ func (b modelBuilder) buildProperty(field reflect.StructField, model *Model, mod
 		jsonName, prop := b.buildStructTypeProperty(field, jsonName, model)
 		return jsonName, modelDescription, prop
 	case fieldKind == reflect.Slice || fieldKind == reflect.Array:
-		jsonName, prop := b.buildArrayTypeProperty(field, jsonName, modelName)
-		return jsonName, modelDescription, prop
+		if fieldType.Elem().Kind() == reflect.Uint8 {
```
liggitt: the special case should probably be inside `buildArrayTypeProperty`

swagger/model_builder.go
```diff
@@ -65,8 +65,12 @@ func (b modelBuilder) addModel(st reflect.Type, nameOverride string) *Model {
 
 	// check for slice or array
 	if st.Kind() == reflect.Slice || st.Kind() == reflect.Array {
-		b.addModel(st.Elem(), "")
-		return &sm
+		if st.Elem().Kind() == reflect.Uint8 {
+			return nil
```
liggitt: is this returning nil because we're representing it as a "string", which is already added as a model? If so, I would put this case up above, right after the isPrimitiveType check, with a really good comment

mikedanese:
Addressed comments.

PR #317, opened 2016-09-01
Adding slice/array functionality to 'type' struct tag.

zhewison:
This pull request was born whilst using a custom type as a slice in a model struct.

The custom type `bson.ObjectId` implements correct marshaling, however when used as a slice/array in a model it does not reflect the type correctly in swagger documentation.

In the swagger docs `[]bson.ObjectId` becomes `[{}]` when it should reflect as `["string"]`

This PR allows you to use `[]` with the `type` tag to override custom types as slices/arrays instead of just basic types.

e.g `type:"[]string"` will result in `["string"]` for your model documentation.


emicklei:
thank you for contributing @zhewison

PR #318, opened 2016-09-08
Change GET to POST for creation

fvm:
Just sayin' :D


emicklei:
this was a test that slipped into one of the previous commits by mistake. Thanks for reporting this.

Issue #306, opened 2016-07-13
restful v2


emicklei:
The idea is to start a new repository (but keeping the package name) that will introduce a new, improved, API to create REST-like applications in Go. With respect to the current status, this will be a major change (hence the v2) as I expect to see a few changes to the current API.

This issue will list the things I like to change/improve
- move Swagger to its own repository allowing for a 2.0 OpenAPI support
- RouteFunction will have a Context parameter (as the last?)
- Replace public vars by functions
- Use CurlyRouter by default
- No caching of request content by default
- Introduce a static container in addition to the current one that allows add/remove WebServices
- Remove request attributes in favor of Context

Other suggestions are welcome.

vikstrous:
My thoughts in no particular order:

I think it's best to stick to the convention of passing a context as the first parameter.

I may be wrong, but I doubt anyone would be upset if the new gorestful 2 supported only swagger 2 and not 1.2.

It would be cool if parts of go-swagger can be reused here, if not all of it.

On July 13, 2016 4:26:02 PM CDT, Ernest Micklei <notifications@github.com> wrote:
>The idea is to start a new repository (but keeping the package name)
>that will introduce a new, improved, API to create REST-like
>applications in Go. With respect to the current status, this will be a
>major change (hence the v2) as I expect to see a few changes to the
>current API.
>
>This issue will list the things I like to change/improve
>
>- move Swagger to its own repository allowing for a 2.0 OpenAPI support
>- RouteFunction will have a Context parameter (as the last?)
>- Replace public vars by functions
>- Use CurlyRouter by default
>- No caching of request content by default
>- Introduce a static container in addition to the current one that
>allows add/remove WebServices
>- Remove request attributes in favor of Context
>
>Other suggestions are welcome.
>
>
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/emicklei/go-restful/issues/306

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

AdoHe:
Really look forward to this, @emicklei could you paste the new repository location here?

emicklei:
fyi, i am working on an initial version that includes the Context. Once all tests pass , I will create the initial commit (which will still be WIP though)

AdoHe:
I think I have time to help with the new version. So really look forward to your init commit.

emicklei:
[...]

PR #319, opened 2016-09-11
no request content cache in v2

AdoHe:
@emicklei According to what we have talked  here #306, I open this PR to disable request content cache, I want to make sure this is just what we want, then I will add an example to show how we can implement the same with filter. PTAL. Besides should we create a new v2 branch, or we just use master branch?

### DIFF
--- a/compress_test.go
+++ b/compress_test.go
@@ -94,7 +94,6 @@ func TestGzipDecompressRequestBody(t *testing.T) {
 	httpRequest.Header.Set("Content-Encoding", "gzip")
 	req.Request = httpRequest
 
-	doCacheReadEntityBytes = false
 	doc := make(map[string]interface{})
 	req.ReadEntity(&doc)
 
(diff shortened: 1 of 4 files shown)


emicklei:
I think it is a good idea to put these changes in its own branch

AdoHe:
@emicklei sorry I am not very clear, what you mean in its own branch?

emicklei:
let's create a branch called "v2" and collect the related changes there

emicklei:
i am going to repeat this change to the current master.

AdoHe:
@emicklei should I open a PR based on master?

emicklei:
@AdoHe I already applied the same changes to the current master. Thanks. Would you be interested in working on swagger2?

AdoHe:
@emicklei I just return from my ooo, will we start working on swagger2 in this year?

emicklei:
yes, swagger2 in on my hot list.  It has been a while since I paid attention to it. Partly because of the decision whether or not to embrace go-swagger pkg.

PR #321, opened 2016-09-18
Accept pointers passed to model builder for swagger

smarterclayton:
From a model perspective we treat `type.Struct` and `*type.Struct`
identically when nested, but if `*type.Struct` is passed to go-restful
via `.Writes()`. Since these are expected to be models, this is more
friendly to integrators.

Added test of passing a struct by pointer.


emicklei:
@smarterclayton thank you for this contribution

PR #325, opened 2016-12-12
Support renaming model type names holistically

smarterclayton:
As schemas become more complicated, the odds of two types in different
packages sharing the same generated "reflect.Type#String()" value
increases (e.g. `pkg/type/v1.Event` and `pkg/history/v1.Event` will both
result in `v1.Event` as a model name). This adds a new optional function
on the swagger.Config that can remap type names before they are written
into the Swagger model. Because types must be referenced across
structures, the mapping needs to happen at all the places that convert
`reflect.Type` to `string`.

Spawned by kubernetes/kubernetes#38071 where we now have duplicate type names.

@mbohlool this will allow us to bypass this problem in the short term.

### DIFF
--- a/swagger/config.go
+++ b/swagger/config.go
@@ -2,6 +2,7 @@ package swagger
 
 import (
 	"net/http"
+	"reflect"
 
 	"github.com/emicklei/go-restful"
 )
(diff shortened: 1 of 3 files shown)


emicklei:
thank you for contributing. 
do you have an example (or the one in kubernetes) for the mapping function?

smarterclayton:
Don't have an example yet, but going to add one.  We're going to convert the Go package name (example, `k8s.io/kubernetes/pkg/apis/meta/v1.Event`) to a shorter version `k8s.kubernetes.apis.meta.v1.Event`.

PR #328, opened 2017-01-30
Additional CORS headers for OPTIONS request

jamesmillerio:
I created an [issue on the Mora project](https://github.com/emicklei/mora/issues/50) when I wasn't able to execute PUT requests due to a CORS violation. The following headers were missing on OPTIONS preflight responses:

- Access-Control-Allow-Headers
- Access-Control-Allow-Methods
- Access-Control-Allow-Origin

Chrome required these and their absense caused the request to get rejected. After digging around and trying to figure out how best to fix it, I came up with a solution in this project that addresses the issue. The tests have been modified to account for the change and verified. Hopefully you'll find it satisfactory. Thanks!

### DIFF
--- a/options_filter.go
+++ b/options_filter.go
@@ -6,21 +6,34 @@ import "strings"
 // Use of this source code is governed by a license
 // that can be found in the LICENSE file.
 
-// OPTIONSFilter is a filter function that inspects the Http Request for the OPTIONS method
+// OptionsFilter is used to create a Filter that implements proper headers for CORS OPTIONS
+// requests. Cross-origin resource sharing (CORS) is a mechanism that allows JavaScript on
+// a web page to make XMLHttpRequests to another domain, not the domain the JavaScript
+// originated from.
+//
+// http://en.wikipedia.org/wiki/Cross-origin_resource_sharing
+// http://enable-cors.org/server.html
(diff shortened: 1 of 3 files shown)


emicklei:
thank you for contributing. would you be so kind as to re-create this PR based on the new repository  called go-restful-swagger12.  I will update the mora repository accordingly.

jamesmillerio:
Sorry if I'm being dense but not sure I follow. You want a PR to add the OptionsFilter struct to go-restful-swagger12? I'm not terribly familiar with go-restful-swagger12 so I wasn't quite sure how those two would go together being from disparate projects.

emicklei:
oops. my bad. I thought it included changes to the swagger project. I should have looked at the diff first.

options_filter.go
```diff
@@ -6,21 +6,34 @@ import "strings"
 // Use of this source code is governed by a license
 // that can be found in the LICENSE file.
 
-// OPTIONSFilter is a filter function that inspects the Http Request for the OPTIONS method
+// OptionsFilter is used to create a Filter that implements proper headers for CORS OPTIONS
```
emicklei: Renaming the type is breaking the api of this package. IMO this must be avoided.
It was spelled in uppercase because it reflects the OPTIONS method.

tekkamanendless:
@jamesmillerio, are you going to make the appropriate changes so that we can get this accepted?

PR #329, opened 2017-02-11
Add metadata to Route

mbohlool:
The PR allows storing metadata with each route. This metadata will not not have any effect on routing or swagger 1.2 spec generation and will be only used by the user of the library to store more information about the route.

### DIFF
--- a/route.go
+++ b/route.go
@@ -34,6 +34,9 @@ type Route struct {
 	ParameterDocs           []*Parameter
 	ResponseErrors          map[int]ResponseError
 	ReadSample, WriteSample interface{} // structs that model an example request or response payload
+
+	// Extra information used to store custom information about the route.
+	Metadata map[string]interface{}
 }
 
 // Initialize for Route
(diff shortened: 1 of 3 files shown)


emicklei:
thank you for contributing. Can you give an example how you are going to use it?
I remember some other issue where it was suggested to provide a reference to the Route from the actual Request. Would that be helpful for your usecase as well?

mbohlool:
We want to provide more information in our spec through extensions about routes. An example would be the object an operation work on and a simpler operation name on it. It would help create fluent interface for our client. Our spec generator relies on go-restful Routes only to generate Operation specs. It made sense to add those extensions to the routes and then reuse them in spec generation.

emicklei:
your spec generator, is that similar to what I am doing at the http://github.com/emicklei/go-restful-openapi project?

mbohlool:
Yes. https://github.com/kubernetes/kubernetes/tree/master/cmd/libs/go2idl/openapi-gen

PR #334, opened 2017-04-10
Update stale comment

gitdr:
Default router was changed to CurlyRouter as part of #306.


emicklei:
thank you for this contribution

PR #339, opened 2017-05-21
Fix commonly misspelled words

spring1843:
Fixes a few typos in comments


emicklei:
many thanks for these fixes!

PR #341, opened 2017-05-24
use golang internal http BasicAuth method

nasermirzaei89:
Golang has a BasicAuth method for HTTP Request. I use it instead of your method.


emicklei:
thanks!

PR #343, opened 2017-06-03
allow user to pass arguments for filter functions

mqliang:


emicklei:
you could stick any option in the request attributes instead

PR #352, opened 2017-09-13
Add support for route condition functions

glb:
Sometimes you need to be able to say that a route only
applies under certain conditions. Now you can.

### DIFF
--- a/jsr311.go
+++ b/jsr311.go
@@ -41,9 +41,29 @@ func (r RouterJSR311) SelectRoute(
 
 // http://jsr311.java.net/nonav/releases/1.1/spec/spec3.html#x3-360003.7.2
 func (r RouterJSR311) detectRoute(routes []Route, httpRequest *http.Request) (*Route, error) {
+	ifOk := []Route{}
+	for _, each := range routes {
+		ok := true
+		for _, fn := range each.If {
+			if !fn(httpRequest) {
+				ok = false
+				break
+			}
+		}
+		if ok {
+			ifOk = append(ifOk, each)
+		}
+	}
+	if len(ifOk) == 0 {
+		if trace {
+			traceLogger.Printf("no Route found (from %d) that passes conditional checks", len(routes))
+		}
+		return nil, NewError(http.StatusNotFound, "404: Not Found")
+	}
+
 	// http method
 	methodOk := []Route{}
-	for _, each := range routes {
+	for _, each := range ifOk {
 		if httpRequest.Method == each.Method {
 			methodOk = append(methodOk, each)
 		}
--- a/route.go
+++ b/route.go
@@ -13,6 +13,11 @@ import (
 // RouteFunction declares the signature of a function that can be bound to a Route.
 type RouteFunction func(*Request, *Response)
 
+// RouteSelectionConditionFunction declares the signature of a function that
+// can be used to add extra conditional logic when selecting whether the route
+// matches the HTTP request.
+type RouteSelectionConditionFunction func(httpRequest *http.Request) bool
+
 // Route binds a HTTP Method,Path,Consumes combination to a RouteFunction.
 type Route struct {
 	Method   string
@@ -21,6 +26,7 @@ type Route struct {
 	Path     string // webservice root path + described path
 	Function RouteFunction
 	Filters  []FilterFunction
+	If       []RouteSelectionConditionFunction
 
 	// cached values for dispatching
 	relativePath string
--- a/route_builder.go
+++ b/route_builder.go
@@ -24,6 +24,7 @@ type RouteBuilder struct {
 	httpMethod  string        // required
 	function    RouteFunction // required
 	filters     []FilterFunction
+	conditions []RouteSelectionConditionFunction
 
 	typeNameHandleFunc TypeNameHandleFunction // required
 
@@ -212,6 +213,21 @@ func (b *RouteBuilder) Filter(filter FilterFunction) *RouteBuilder {
 	return b
 }
 
+// If sets a condition function that controls matching the Route based on custom logic.
+// The condition function is provided the HTTP request and should return true if the route
+// should be considered.
+//
+// Efficiency note: the condition function is called before checking the method, produces, and
+// consumes criteria, so that the correct HTTP status code can be returned.
+//
+// Lifecycle note: no filter functions have been called prior to calling the condition function,
+// so the condition function should not depend on any context that might be set up by container
+// or route filters.
+func (b *RouteBuilder) If(condition RouteSelectionConditionFunction) *RouteBuilder {
+	b.conditions = append(b.conditions, condition)
+	return b
+}
+
 // If no specific Route path then set to rootPath
 // If no specific Produces then set to rootProduces
 // If no specific Consumes then set to rootConsumes
@@ -254,6 +270,7 @@ func (b *RouteBuilder) Build() Route {
 		Consumes:       b.consumes,
 		Function:       b.function,
 		Filters:        b.filters,
+		If:             b.conditions,
 		relativePath:   b.currentPath,
 		pathExpr:       pathExpr,
 		Doc:            b.doc,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 Change history of go-restful
 =
+2017-09-13
+- added route condition functions using `.If(func)` in route building.
+
 2017-02-16
 - solved issue #304, make operation names unique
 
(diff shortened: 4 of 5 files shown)


emicklei:
interesting feature. do you have real examples of conditions that you have implemented for your services?

glb:
Thanks! I was thinking that this PR probably needed some more clarifying commentary.

I'm a fan of using HTTP content negotiation where possible for versioning my entity representations, so I like to use things like `application/vnd.com.example.entity.v1+json`. However, I have learned that many many folks really really prefer to just use `application/json`, so I'd like to support that.

Also, I've seen some really amazing API versioning examples from folks like the ones at Stripe ([link](https://stripe.com/blog/api-versioning)) who latch users into the API version that they're using when they start, and let them keep using that version until they're ready to update.

As a result, I'm hoping to provide a compromise solution that does something like the following:
```
ws.Route(ws.
	GET("/resources").
	To(listResourcesV2).
	If(func(request *http.Request) bool {
		// MimeTypeV2 was introduced in the 2017-09-08 API version
		return strings.Compare(apiVersion.FromHTTPRequest(request), "2017-09-08") >= 0
	}).
	Produces(MimeTypeV2, restful.MIME_JSON))

ws.Route(ws.
	GET("/resources").
	To(listResourcesV1).
	If(func(request *http.Request) bool {
		// MimeTypeV1 was introduced in the 2017-08-17 API version
		return strings.Compare(apiVersion.FromHTTPRequest(request), "2017-08-17") >= 0
	}).
	Produces(MimeTypeV1, restful.MIME_JSON))
```
(there are less-verbose ways to write the above but for the purposes of this example I've left it verbose)

The key part that the `If` condition lets me do is return 415 / 406 properly if the person is using the wrong MIME type for their API version. I had previously just registered the route saying that it produced / consumed both versioned MIME types (e.g. `Produces(MimeTypeV2, MimeTypeV1, restful.MIME_JSON)`) but that was pretty ugly and didn't work very well.

I have another idea in mind for implementing entity migration like what the Stripe folks did so that I don't have to have separate `listResourcesV1` and `listResourcesV2` functions, but I'll bring that up in another PR once it's had a little more time to mature.

Thoughts?

emicklei:
Thank for providing the examples and your considerations. I understand your usecase. 
I did not expect that having multiple produces would produce (too) complex code.

My initial idea would be something like

```
func (s Service) listResources(request *restful.Request, response *restful.Response) {
    if strings.Compare(apiVersion.FromHTTPRequest(request.Request), "2017-09-08") >= 0 {
           s.listResourcesV2(request,response)
           return
    }
    if strings.Compare(apiVersion.FromHTTPRequest(request.Request), "2017-08-17") >= 0 {
            s.listResourcesV1(request,response)
            return
    }
    response.WriteHeader(415)
}
```

glb:
Thanks!

The example you provided is similar to what I'd originally tried, but I found that there was some difficulty / complexity pushed into the service methods when trying to get the correct HTTP responses. For example, consider that it's really quite nice for the developer to be able to use `response.WriteHeaderAndEntity(responseCode, responseEntity)`, which uses `response.EntityWriter()` to figure out which media type matches the route and the HTTP request headers best, as in:
```
func (s Service) listResourcesV1(request *restful.Request, response *restful.Response) {
   // ... construct list of resources
   response.WriteEntity(resources)
}
```
and they don't worry about the MIME stuff at all. This leverages the super-helpful negotiation code that `go-restful` already has.

However, with this approach, if the route produces `MimeTypeV2`, `MimeTypeV1`, and `restful.MIME_JSON`, then there is strangeness for the following cases:
* user is at API version 1 but doesn't provide an `Accept` header: they get `Content-Type: MimeTypeV2` but should get `Content-Type: MimeTypeV1`.
* user is at API version 2 but sends an `Accept` header with `MimeTypeV1` (for example if they update their provisioning but have an old API client that they forgot about): they get `Content-Type: MimeTypeV1` but should get a `406 Not Acceptable` response.

There are symmetrical issues on the `Consumes` side.

All of these issues _can_ be dealt with in the service methods or some wrapper, _I think_, but with some difficulty as they have to replicate the logic in `response.EntityWriter()`.

I found it much nicer to have `EntityWriter` continue to do the great job it does to make sure that all the HTTP content negotiation happens correctly for the route and keep the service logic simple. It's also easy to make the route registration more data-driven than I showed above so that I can encapsulate the changes per API version in a clear and simple way.

I'm pretty focused on API versioning at the moment, so I haven't put deep consideration into other use cases for the changes in this PR, but it seems like the `If` condition would be helpful in creating an API that's responsive to context in other useful ways. I'm doing a fair amount with filters as well, but this seemed to be a capability that needed to be done slightly differently.

emicklei:
ok, now I understand how this can be more complex with mixed access patterns. (btw, at our company we decided to put the major version indicator in the path, e.g. /v1/accounts/{id} so we can drop the whole resource at some point).

So, let's add this feature which slightly affects the route detection logic. I plan to write some comments on the suggested implementation. Please have a look at those.

jsr311.go
```diff
@@ -41,9 +41,29 @@ func (r RouterJSR311) SelectRoute(
 
 // http://jsr311.java.net/nonav/releases/1.1/spec/spec3.html#x3-360003.7.2
 func (r RouterJSR311) detectRoute(routes []Route, httpRequest *http.Request) (*Route, error) {
+	ifOk := []Route{}
```
emicklei: Did you consider combining these two loops over the routes?
Looks like the method detection is independent of the IfFunction.

route.go
```diff
@@ -21,6 +21,7 @@ type Route struct {
 	Path     string // webservice root path + described path
 	Function RouteFunction
 	Filters  []FilterFunction
+	If       []RouteIfFunction
```
emicklei: I hesitate whether the name RouteIfFunction is the best fit. Is it clear enough to what process the "If" is related?
Its behaviour is more like a RouteSelectionConditionFunction.

routeif.go
```diff
@@ -0,0 +1,5 @@
+package restful
+
+import "net/http"
+
+type RouteIfFunction func(httpRequest *http.Request) bool
```
emicklei: Did you consider using the wrapper type as argument ?
I know the detectRoute does not have this wrapper but what if the condition needs a path parameter value?

jsr311.go
glb: I noticed that you've got a few passes in this method that iteratively filter through the routes and produce `methodOk`, `mediaOk`, and `inputMediaOk`, so I just followed that pattern and didn't consider something else.

route.go
glb: I'd be happy to rename. I think that the builder reads better as `.If(...)` but changing the type to `RouteSelectionConditionFunction` does make the purpose and usage clearer.

routeif.go
glb: I considered it but wasn't sure how to get / provide the wrapper object in that context, and since my conditions don't need them I didn't put a lot of time into it. How would one do that? Would it be much additional re-parsing?

glb:
Thanks for the consideration and comments! I have added a couple of replies with questions, once I understand the next steps I'll make the changes.

jsr311.go
emicklei: yes, ok leave it like this. maybe we could optimize later.  maybe put de method check (is cheap) before the conditional function loop.

route.go
emicklei: The "If" is ok, it's readable like this.  Please rename the type.

routeif.go
emicklei: I had a quick look and get the feeling it will complicate/change the implementation considerably. Let's leave it like this.

route_builder.go
```diff
@@ -24,6 +24,7 @@ type RouteBuilder struct {
 	httpMethod  string        // required
 	function    RouteFunction // required
 	filters     []FilterFunction
+	iffunctions []RouteSelectionConditionFunction
```
glb: I'm thinking I should rename this to `conditions`; it won't change the external interface but seems like it would read better.

route_builder.go
```diff
@@ -24,6 +24,7 @@ type RouteBuilder struct {
 	httpMethod  string        // required
 	function    RouteFunction // required
 	filters     []FilterFunction
+	conditions []RouteSelectionConditionFunction
```
glb: I renamed this from `iffunctions` to `conditions` as that seemed to read better.

glb:
OK, I think I've made the changes and have cleaned up a bit, good to go?

emicklei:
thank you for adding this feature and discussing the solution. i will create a new tag on the repo.

glb:
Thank you!

Issue #354, opened 2017-09-19
Response does not implement http.Hijacker


johnweldon:
The go-restful `Response` type does not implement `http.Hijacker`, and the underlying `http.ResponseWriter` loses the potential `http.Hijacker` implementation when it gets stored as an embedded interface, making it impossible to use a type assertion on the `Response` to get a `Hijacker`, even if the underlying `ResponseWriter` actually does implement `Hijacker`.

johnweldon:
Incidentally related to #338 

This feature is needed to allow the gorilla websocket implementation to work with go-restful.

emicklei:
thank you for contributing

PR #355, opened 2017-09-19
Expose http.Hijacker interface on Response

johnweldon:
Addresses #354

### DIFF
--- a/entity_accessors_test.go
+++ b/entity_accessors_test.go
@@ -49,7 +49,7 @@ func TestKeyValueEncoding(t *testing.T) {
 	// Write
 	httpWriter := httptest.NewRecorder()
 	//								Accept									Produces
-	resp := Response{httpWriter, "application/kv,*/*;q=0.8", []string{"application/kv"}, 0, 0, true, nil}
+	resp := Response{ResponseWriter: httpWriter, requestAccept: "application/kv,*/*;q=0.8", routeProduces: []string{"application/kv"}, prettyPrint: true}
 	resp.WriteEntity(b)
 	t.Log(string(httpWriter.Body.Bytes()))
 	if !kv.writeCalled {
--- a/response.go
+++ b/response.go
@@ -5,7 +5,9 @@ package restful
 // that can be found in the LICENSE file.
 
 import (
+	"bufio"
 	"errors"
+	"net"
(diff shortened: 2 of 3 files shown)


response.go
```diff
@@ -19,17 +21,19 @@ var PrettyPrintResponses = true
...
-	contentLength int      // number of bytes written for the response body
-	prettyPrint   bool     // controls the indentation feature of XML and JSON serialization. It is initialized using var PrettyPrintResponses.
-	err           error    // err property is kept when WriteError is called
+	requestAccept string        // mime-type what the Http Request says it wants to receive
+	routeProduces []string      // mime-types what the Route says it can produce
+	statusCode    int           // HTTP status code that has been written explicitly (if zero then net/http has written 200)
+	contentLength int           // number of bytes written for the response body
+	prettyPrint   bool          // controls the indentation feature of XML and JSON serialization. It is initialized using var PrettyPrintResponses.
+	err           error         // err property is kept when WriteError is called
+	hijacker      http.Hijacker // if underlying ResponseWriter supports it
```
emicklei: is adding a field necessary? the Hijack() function can access the embbed ResponseWriter.

entity_accessors_test.go
```diff
@@ -49,7 +49,7 @@ func TestKeyValueEncoding(t *testing.T) {
 	// Write
 	httpWriter := httptest.NewRecorder()
 	//								Accept									Produces
-	resp := Response{httpWriter, "application/kv,*/*;q=0.8", []string{"application/kv"}, 0, 0, true, nil}
+	resp := Response{ResponseWriter: httpWriter, requestAccept: "application/kv,*/*;q=0.8", routeProduces: []string{"application/kv"}, prettyPrint: true}
```
emicklei: thank you for making these changes;  at the time of initial writing, I was lazy and then the arguments started growing...

response.go
johnweldon: The problem is that the embedded ResponseWriter is stored as the interface, and you lose access to the implementation that supports the `Hijack()` method. (You can see this by attempting to do a type assertion on the embedded ResponseWriter as a Hijacker - it will fail.

emicklei: thanks for clarifying this.

Issue #27, opened 2013-07-14
How to enable CORS


the42:
in net/http I would do the following:

```
if origin := req.Header.Get("Origin"); origin != "" {
    w.Header().Set("Access-Control-Allow-Origin", origin)
}
```

Where is the right place to enable CORS in go-restful? I would assume a filter ...

yasser1984ir:
Isn't this how you would do the same?

https://github.com/emicklei/go-restful/blob/master/examples/restful-CORS-filter.go

emicklei:
i think the OPTIONSFilter is no longer needed (in the example). Have to verify that.

yasser1984ir:
This ultimately worked for me, and OPTIONSFilter was unnecessary.

```go
	cors := restful.CrossOriginResourceSharing{
		ExposeHeaders:  []string{"X-My-Header"},
		AllowedHeaders: []string{"Content-Type", "Accept"},
		AllowedDomains: []string{FrontEndDomain},
		CookiesAllowed: false,
		Container:      wsContainer}
	wsContainer.Filter(cors.Filter)
```

PR #357, opened 2017-10-02
Add Deprecate method to RouteBuilder 

schleppy:
This PR puts adds a Deprecated() method to the RouteBuilder.  The deprecated value in turn is set on the Route which go-restful-openapi can use to set the spec.Operation's Deprecated property which is then called out in the swagger UI.

Additional PR for go-restful-openapi will come as well to close the loop.

### DIFF
--- a/route.go
+++ b/route.go
@@ -43,6 +43,9 @@ type Route struct {
 
 	// Extra information used to store custom information about the route.
 	Metadata map[string]interface{}
+
+	// marks a route as deprecated
+	Deprecated bool
 }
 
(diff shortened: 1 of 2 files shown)


emicklei:
thank you for contributing

schleppy:
No problem, happy to help.  Any chance of go-restful-openapi [#27](getting https://github.com/emicklei/go-restful-openapi/pull/27) to rebuild?

vikstrous:
Just found this. This is awesome! Thanks!

PR #358, opened 2017-10-03
Add CollectionFormat to Parameter type.  

schleppy:
When used with `AllowMultiple(true)` this will trigger the swagger UI to build the appropriate entry component.  

For example:

image

An additional change is required in go-restful-openapi to populate the appropriate pieces.  I will send a PR for that if this gets merged.

### DIFF
--- a/web_service.go
+++ b/web_service.go
@@ -118,7 +118,7 @@ func (w *WebService) QueryParameter(name, description string) *Parameter {
 // QueryParameter creates a new Parameter of kind Query for documentation purposes.
 // It is initialized as not required with string as its DataType.
 func QueryParameter(name, description string) *Parameter {
-	p := &Parameter{&ParameterData{Name: name, Description: description, Required: false, DataType: "string"}}
+	p := &Parameter{&ParameterData{Name: name, Description: description, Required: false, DataType: "string", CollectionFormat: CollectionFormatCSV.String()}}
 	p.beQuery()
 	return p
 }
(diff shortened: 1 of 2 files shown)


timothyberger:
+1

marcusking01:
+1

bipuldutta:
+1

web_service.go
```diff
@@ -118,7 +118,7 @@ func (w *WebService) QueryParameter(name, description string) *Parameter {
 // QueryParameter creates a new Parameter of kind Query for documentation purposes.
 // It is initialized as not required with string as its DataType.
 func QueryParameter(name, description string) *Parameter {
-	p := &Parameter{&ParameterData{Name: name, Description: description, Required: false, DataType: "string"}}
+	p := &Parameter{&ParameterData{Name: name, Description: description, Required: false, DataType: "string", CollectionFormat: CollectionFormatMulti.String()}}
```
schleppy: I added a default CollectionFormat here only because it has no effect if AllowMultiple is not set to true.  The alternative is to handle the existence of a non-empty string value in the build_path.go of go-restful-openapi.

emicklei: The [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) says ```Default value is csv``` so maybe we should use this instead.

schleppy: Ah, good call.  I will change that.  Do you prefer a new squashed CR, or are you ok with multiple commits?

emicklei: squash it yes please, i will merge right after.

schleppy:
@emicklei all ready.  I will have a PR on the go-restful-openapi as well coming after this gets merged.

emicklei:
thank you for this contribution. awaiting the openapi PR

Issue #363, opened 2018-01-19
PathParameters inconsistent with jsr311 router when route has greedy parameter


iainduncani:
In the situation where there are two routes:

```
/resource/{name:*}
/resource/{name:*}/child
```

If using the jsr311 router to get `/resource/wibble/fish/child` the jsr311 router correctly picks the second route as the handler, however the name path parameter is set to `wibble/fish/child`.

This is because the route handler for the path parameter does the same thing whether the curly or jsr311 router is being used:

https://github.com/emicklei/go-restful/blob/master/route.go#L141

I think the solution would be to make the route.go accept the parameter processing function when it is created and the existing code moved into curly.go.  jsr311.go could then provide an alternative behaviour that is consistent with it's routing policy.

emicklei:
thank you for sharing this. Would you care to work on such proposal together?

iainduncani:
Yes, I'm happy to put something together, mostly wanted to use the issue to see if this is something you would consider merging if I looked at the implementation!

iainduncani:
I've done a bit more digging and I don't think this will be too hard to get out of the JSR311 route as the way the regexp is built up means we have all the information in the thing returned from the regexp matcher on this line:

https://github.com/emicklei/go-restful/blob/master/jsr311.go#L127

One thing I'm not sure about though is where to put this information to not end up processing the regex again.  It looks like the route objects that are created are created when they are defined so there isn't one-per request that is built up by the jsr311router like I had hoped.  I guess it's not the end of the world to re-process the regex but it seems like it would be nicer to just set the path params onto the route when we do this selecting.

iainduncani:
Hi @emicklei I've just raised a PR for this, let me know if you are happy with the changes!

iainduncani:
Hi @emicklei I've added a new fix that doesn't change any of the existing interfaces so hopefully it is more appropriate

emicklei:
can we close this?

iainduncani:
Yes! Thanks for you help merging the fix.

PR #365, opened 2018-01-25
Jsr311 path param processing

iainduncani:
Fix for issue #363

### DIFF
--- a/curly.go
+++ b/curly.go
@@ -5,6 +5,7 @@ package restful
 // that can be found in the LICENSE file.
 
 import (
+	"bytes"
 	"net/http"
 	"regexp"
 	"sort"
@@ -18,7 +19,7 @@ type CurlyRouter struct{}
 // for the WebService and its Route for the given Request.
 func (c CurlyRouter) SelectRoute(
 	webServices []*WebService,
-	httpRequest *http.Request) (selectedService *WebService, selected *Route, err error) {
+	httpRequest *http.Request) (selectedService *WebService, selected *Route, params map[string]string, err error) {
 
 	requestTokens := tokenizePath(httpRequest.URL.Path)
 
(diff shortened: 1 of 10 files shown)


curly.go
```diff
@@ -18,7 +19,7 @@ type CurlyRouter struct{}
 // for the WebService and its Route for the given Request.
 func (c CurlyRouter) SelectRoute(
 	webServices []*WebService,
-	httpRequest *http.Request) (selectedService *WebService, selected *Route, err error) {
+	httpRequest *http.Request) (selectedService *WebService, selected *Route, params map[string]string, err error) {
```
emicklei: this is a breaking change as the exposed function signature is changed.  was this really needed?  I will continue reviewing.

curly_test.go
```diff
@@ -158,6 +158,63 @@ func TestExtractParameters_Wildcard3(t *testing.T) {
 	}
 }
 
+func TestExtractParameters_Wildcard4(t *testing.T) {
```
emicklei: why did you move the function definition to another file?

curly.go
```diff
@@ -162,3 +168,46 @@ func (c CurlyRouter) computeWebserviceScore(requestTokens []string, tokens []str
 	}
 	return true, score
 }
+
+// Extract the parameters from the request url path
+func extractParameters(r *Route, urlParts []string) map[string]string {
```
emicklei: why did you move this definition to another file?

curly.go
iainduncani: Ah, I hadn't appreciated that other people might supply their own router so I thought as long as I updated the two in here this would be OK.  I'd done this as I wanted the router to be responsible for doing the calculation of the parameters as it's them that changes how they do routing and handling of the params.  Instead I think it's probably better to do this through an extra interface that the routers can optionally implement, it's not quite as clean but better than introducing a breaking change...

curly.go
iainduncani: To put it into the router as different routers treat params differently in their routing handling, as with the comment above I'll use this default behaviour if the router doesn't calculate the params itself

iainduncani:
Closed to make extra changes that don't introduce breaking change

PR #366, opened 2018-01-26
Make JSR 311 routing and path param processing consistent

iainduncani:
Fix for issue #365


emicklei:
thank for contributing!.

The CurlyRouter was created to optimize for the case that a path does not have regexes. Does this change causes the Curly router to behave differently w.r.t path processing?

iainduncani:
No, the curly router should behave exactly the same

PR #376, opened 2018-06-01
Add a new method Request.QueryParameters() to retrieve multiple values

moutoum:
This functions returns all the values for a query parameter name. It's working with the `foo=bar&foo=baz` format.
For example, for the previous example, req.QueryParameters("foo") will return `[bar baz]`
The function does not use the CollectionFormat field in the Parameter type, because parameters are not reachable in the portion of the code.

The function does not use the CollectionFormat field in the Parameter type, because parameters are not reachable in the portion of the code.


emicklei:
thanks. can you add a test for this?

moutoum:
No problem, I added the test

emicklei:
thank you for contributing

PR #379, opened 2018-06-30
add json-iterator for go-restful

iyacontrol:
[json-iterator](https://github.com/json-iterator/go) is a high-performance 100% compatible drop-in replacement of "encoding/json". some go web framework already supported it, for example gin and iris etc.

use encoding/json as default json package but you can change to jsoniter by build from other tags
`go build -tags=jsoniter .`

for some api application,this is necessary.

### DIFF
--- a/entity_accessors.go
+++ b/entity_accessors.go
@@ -5,9 +5,7 @@ package restful
 // that can be found in the LICENSE file.
 
 import (
-	"encoding/json"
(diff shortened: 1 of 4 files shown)


emicklei:
thank you for adding this feature. I like the solution that is it controlled by a build flag ; it does not introduce a dependency if users of this package stick to the standard json pkg.

PR #382, opened 2018-07-19
log: Deprecate in favor of github.com/go-log/log

wking:
Builds on #381; review that first.

go-restful's Logger interface is from b80aca09 (#188), at which point the interface included both the `Print` and `Fatal` method families.  That interface moved to the log subpackage in 18c1d52a (#192).  And the interface was reduced to just `Print` and `Printf` in a40a6af9 (#194), but the interface is clearly not go-restful-specific.

More recently, [github.com/go-log/log][1] ([godocs][2]) was created as a stand-alone package with a similar interface (it has `Log` and `Logf` instead of `Print` and `Printf`).  Just this morning, go-log grew [a subpackage to convert from `Print`/`Printf` to `Log`/`Logf`][3].  That means we can offload our generic-logger shim to go-log, save a few lines of generic code here (after removing the deprecated code), and automatically benefit from go-logs other implementations (e.g. you can use [`info.New`][4] to plug in a logger based on [glog's `Verbose`][5]).

I've deprecated the old interfaces, but kept them around to facilitate migrations to the new approach.  The only breaking change should be the removal of `log.Logger`, so that folks who are setting that directly (vs. going through one of the two `SetLogger` functions) will get compile-time errors and notice that they have to update their usage.  If we don't mind jumping through a few extra hoops, we could keep `log.Logger` around for now and stage migration with:

Phase 1:

* Replace the public `restful.Logger` with a `restful.SetLogLogger` setter, that takes a go-log (`Log`/`Logf`) interface.
* Calling `restful.SetLogger`, `restful.SetLogLogger`, or `log.SetLogger` will all set `log.Logger`.  We'd need a `Log`/`Logf` -> `Print`/`Printf` shim for `SetLogLogger`.
* Keep importing go-restful/log and using `log.Print` and `log.Printf` where I'm currently using `Logger.Log` and `Logger.Logf`.

Phase 2:

* Drop go-restful/log and the other interfaces deprecated in this commit.
* Replace the removed `log.Logger` global with a new `restful.Logger`.
* Replace the `log.Print` and `log.Printf` calls with `Logger.Log` and `Logger.Logf`.
* Deprecate `SetLogLogger`.

Phase 3:

* Drop `SetLogLogger`.

Alternatively, you could make the global private (`restful.logger`) and keep `SetLogLogger`.  But I expect the go-log `Logger` interface to be stable enough that we don't need to hide it behind a setter, expecially one with as ugly a name as `SetLogLogger`, or `SetLogger2`, or whatever ;).  So my first take at this PR just drops `log.Logger` immediately to set us up for a simpler, two-phase migration (where phase 2 is just "drop all the deprecated stuff").

I also see that there's a [move-log branch][6] (without an associated PR) doing some similar log-subpackage shuffling.  As of d8185d6c, that branch is removing the log subpackage entirely, so maybe my backwards compat concerns are overblown ;).

[1]: https://github.com/go-log/log
[2]: https://godoc.org/github.com/go-log/log
[3]: https://godoc.org/github.com/go-log/log/print#New
[4]: https://godoc.org/github.com/go-log/log/info#New
[5]: https://godoc.org/github.com/golang/glog#Verbose
[6]: https://github.com/emicklei/go-restful/tree/move-log

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,7 @@ language: go
 go:
   - 1.x
 
+before_install:
+  - go get -d github.com/go-log/log
+
 script: go test -v
--- a/container.go
+++ b/container.go
@@ -97,7 +97,7 @@ func (c *Container) Add(service *WebService) *Container {
 	// cannot have duplicate root paths
 	for _, each := range c.webServices {
 		if each.RootPath() == service.RootPath() {
-			log.Printf("[restful] WebService with duplicate root path detected:['%v']", each)
+			log.Printf("WebService with duplicate root path detected:['%v']", each)
 			os.Exit(1)
 		}
 	}
@@ -139,7 +139,7 @@ func (c *Container) addHandler(service *WebService, serveMux *http.ServeMux) boo
 
 func (c *Container) Remove(ws *WebService) error {
 	if c.ServeMux == http.DefaultServeMux {
-		errMsg := fmt.Sprintf("[restful] cannot remove a WebService from a Container using the DefaultServeMux: ['%v']", ws)
+		errMsg := fmt.Sprintf("cannot remove a WebService from a Container using the DefaultServeMux: ['%v']", ws)
 		log.Print(errMsg)
 		return errors.New(errMsg)
 	}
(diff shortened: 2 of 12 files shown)


emicklei:
thank you for putting time in this improvement and PR. I need a few days to find time to investigate this as I am now on some holidays in Austria.

emicklei:
hi,  i looked at this PR again (far too late) and realised that the whole logging in Go is just a mess. 
https://www.reddit.com/r/golang/comments/6bptkl/why_isnt_there_a_logger_interface_in_the_standard/

So many packages around that try to solve it and still there is a new one, go-log/log.  The problem with this package is that it brings far too many dependencies (logrus etc). This PR will eventually have breaking changes. I do not see the benefit of bringing this in.

emicklei:
i am closing this now as people show no interest. feel free to re-open later

PR #386, opened 2018-12-11
Add support for disabling HTML escaping behavior in JSON encoder

imjoey:
Currently there is no convenient way to disable HTML escaping for JSON rendering. 

By default, both `encoding/json` and `jsoniter-go` libraries are going to escape `<`, `>` and `&` to keep some browsers from misinterpreting JSON output as HTML. While this default behavior is not suitable for retrieving URLs which use the ampersand `&`to separate multiple query parameters. So the URL in JSON strings `http://www.google.com/search?q=foo&q=bar` will be escaped to `http://www.google.com/search?q=foo\u0026q=bar`. 

So as to provide an easy way to solve this, some changes are proposed in this pr:
* Add a global variable `JSONEscapeHTMLResponses` with a default `true` value to control the HTML escaping behavior of all the reponses
* Add a property named `jsonEscapeHTML` (as well as its setter method) in `Response` for per response behavior controlling
* Add the corresponding logic in `writeJSON` method

### DIFF
--- a/entity_accessors.go
+++ b/entity_accessors.go
@@ -144,19 +144,14 @@ func writeJSON(resp *Response, status int, contentType string, v interface{}) er
 		// do not write a nil representation
 		return nil
 	}
+	enc := NewEncoder(resp)
 	if resp.prettyPrint {
-		// pretty output must be created and written explicitly
-		output, err := MarshalIndent(v, "", " ")
-		if err != nil {
-			return err
-		}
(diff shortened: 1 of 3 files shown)


imjoey:
Well, according to the failed CI testing case, the method `json.MarshalIndent` for pretty printing should be kept, while it dose not support the HTML escaping option. hmmm, close this for now.

PR #387, opened 2019-01-21
simplify inputMediaOk var assignment

Quasilyte:
This avoids suspicious re-assignment to a variable.

Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>


emicklei:
thank you for fixing this

Issue #389, opened 2019-02-01
Double compression 


bgadrian:
Hello,

I have a small issue/question, how can I disable the gzip compression (from container.go) for a specific response.

Sometimes I have a content that is already compressed, and I do not want to decompress and let the restful compress it again, for optimization purposes.

I have found the contentEncodingEnabled flag but that is for all the routes and requests, I want to have a more granular effect.

Maybe if the container will look if a ContentEncoding header already exists? And do not compress it then?

emicklei:
hi, i understand your usecase. Maybe indeed you should be able to control this per response. 
Will investigate teh ContentEncoding suggestion

emicklei:
Alternatively, we could have an extra field in the the restful.Response to indicate that compression must be skipped.

bgadrian:
Thanks for the answer!

From what I saw, the first problem in any fix is that the container compression stream is started before executing the handler, so without any major rehaul I guess the only way it would be to flag/filter only a specific route, which would be acceptable for most use cases I guess.

emicklei:
yes, that is done because any filter should be able to abort the chain and respond by itself.

Furthermore, the NewCompressingResponseWriter already write the content-encoding header upon creation.

What if you create a separate container, disable contentEncoding and set the ServeMux to use the Default. Then you have 2 containers, one with, one without encoding. Then for each WebService choose the right container.

emicklei:
or func (r *Route) EnableContentEncoding(enabled bool){}

bgadrian:
That sounds better, but where to keep the flag, in the Routes metadata? 

I can make a PR if is ok

emicklei:
or just a field in Route, same as WebService.  

yes, PR is welcome

emicklei:
i think this can be closed.

bgadrian:
The functionality is there, but I forgot to
* add the functionality in the Route Builder
* add tests

PR #390, opened 2019-02-10
#389 the route can now override the container.contentEncodingEnabled flag

bgadrian:
If set, the new route flag will override the compression setting used by the container. 

It solved the double-compression issue, or you can add auto compression just for specific routes.

### DIFF
--- a/container.go
+++ b/container.go
@@ -220,20 +220,6 @@ func (c *Container) dispatch(httpWriter http.ResponseWriter, httpRequest *http.R
 		}()
 	}
 
-	// Detect if compression is needed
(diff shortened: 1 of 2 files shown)


emicklei:
thank you contributing. i created a new tag v2.9.0 using these commits.

bgadrian:
My pleasure.

I realized that I forgot to maintain the Chain pattern, the new Route method does not return the route instance.

It is consistent with the container.encoding method but it is not with the other route methods.

emicklei:
ai, missed that. will update it soon

PR #395, opened 2019-03-26
Reduce allocations in per-request methods to improve performance

smarterclayton:
As part of performance evaluations in Kubernetes go-restful was a
significant source of time and more importantly memory allocations
during request processing. Profiling uncovered a number of places
where some specific changes could be made to avoid allocating which
reduces CPU use during garbage collection on per-request actions.

The following three changes were the smallest / easiest and should
be relatively easy to review. There are other hot spots that still
need to be investigated which may require more aggressive changes.

#### Reduce allocations in route selection

The use of new arrays for each phase of detectRoute can be simplified: each
pass filters the existing array and whatever is left at the end is
returned. While somewhat harder to read, this method is a significant 
source of allocations.

#### Reduce allocations in Accept and Content-Type matching

The mime handling methods can do a single pass over the header without 
needing to allocate. This improves performance of these methods in higher
traffic API servers.

#### Prevent additional allocations when sorting routes for selection

Use of an uninitialized array and the use of sort.Reverse cause extra
allocations in request routing which are unnecessary.

### DIFF
--- a/curly.go
+++ b/curly.go
@@ -45,14 +45,14 @@ func (c CurlyRouter) SelectRoute(
 
 // selectRoutes return a collection of Route from a WebService that matches the path tokens from the request.
 func (c CurlyRouter) selectRoutes(ws *WebService, requestTokens []string) sortableCurlyRoutes {
-	candidates := sortableCurlyRoutes{}
+	candidates := make(sortableCurlyRoutes, 0, 8)
 	for _, each := range ws.routes {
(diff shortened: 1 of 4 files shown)


smarterclayton:
@wojtek-t FYI

emicklei:
@smarterclayton thank you for your effort and work to improve the performance.

To be honest, I did not spent time on measurements to see where it can be improved.

I will review your changes carefully.

PR #412, opened 2019-08-31
fix compile error for example/restful-serve-static.go

00arthur00:
missing log package, which contributes to a compile error


emicklei:
thanks

Issue #394, opened 2019-03-20
Support google "Custom Methods" style urls


sindrepm:
Does `go-restful` currently have support for routes like this?

`ws.Route(ws.POST("/projects/{name}:customVerb").To(h.handleCustomVerb))`

For example Google uses this pattern to implement custom methods on resources.
https://cloud.google.com/apis/design/custom_methods

I tried but just get a 404 error when hitting the endpoint.

emicklei:
hi, the closest setup you can do now is similar to https://github.com/emicklei/go-restful/blob/master/examples/restful-path-tail.go

However, the concept is interesting and I would like to see what is needed to support this.

emicklei:
so
```
func main() {
	ws := new(restful.WebService)
	ws.Route(ws.GET("/{name:*}").To(hello))
	restful.Add(ws)
	log.Fatal(http.ListenAndServe(":8080", nil))
}

func hello(req *restful.Request, resp *restful.Response) {
	io.WriteString(resp, req.PathParameter("name"))
}
```
with

    curl http://localhost:8080/name:customVerb

gives you:

    name:customVerb

sindrepm:
> so
> 
> ```
> func main() {
> 	ws := new(restful.WebService)
> 	ws.Route(ws.GET("/{name:*}").To(hello))
> 	restful.Add(ws)
> 	log.Fatal(http.ListenAndServe(":8080", nil))
> }
> 
> func hello(req *restful.Request, resp *restful.Response) {
> 	io.WriteString(resp, req.PathParameter("name"))
> }
> ```
> with
> 
> ```
> curl http://localhost:8080/name:customVerb
> ```
> gives you:
> 
> ```
> name:customVerb
> ```

That would require all custom verbs to be handled by the same route / passed to the same handler.

emicklei:
yes, i realise that's not want you want. 
a better suggestion is create a new CustomMethodsRouter, using the CurlyRouter as an example.
maybe it is only the `tokenizePath` that needs to be aware of this mode of operation. If that turns out to be the solution , we could extend CurlyRouter with this configurable behaviour.

emicklei:
@sindrepm did you find a solution to this issue?

ShiningRush:
@sindrepm @emicklei  i have already support google custom verb, plz check the pr

emicklei:
this feature has been implemented by @ShiningRush

Issue #374, opened 2018-05-24
Go module support


ulrichSchreiner:
hi,

i know that vgo is not released. but it would be great to support it in the near future. with the current version of `vgo` you cannot import a version-2 of `go-restful`. the [design document](https://research.swtch.com/vgo-module) says:
>  Again, the go.mod file is required for v2 and later so that vgo can use the module line as a sign that the code has been written with semantic import versioning in mind, ...

to support multiple versions, the packages should have the version in their import path (`.../v2`).

so the author of `vgo` want the version 2 (or later versions) to have a `go.mod` file. this is not present in the current code. this leads to the following situation:
~~~
$ vgo list -t github.com/emicklei/go-restful   
github.com/emicklei/go-restful
        v1.0.0
        v1.0.1
        v1.1.0
        v1.1.1
        v1.1.2
        v1.1.3
~~~
Great!
~~~
$ vgo list -t github.com/emicklei/go-restful/v2
github.com/emicklei/go-restful/v2
        v2.2.1
        v2.3.0
        v2.4.0
        v2.5.0
        v2.6.0
        v2.6.1
        v2.7.0
~~~
Great!

But:
~~~
$ vgo get github.com/emicklei/go-restful/v2       
vgo: finding github.com/emicklei/go-restful/v2 v2.7.0
vgo: github.com/emicklei/go-restful/v2 v2.7.0: missing go.mod
vgo: finding github.com/emicklei/go-restful/v2 v2.7.0
vgo: github.com/emicklei/go-restful/v2 v2.7.0: missing go.mod
vgo get: missing go.mod
~~~
:-(
Here `vgo` complains about the missing `go.mod` file which indicates that the module does semantic versioning.

So adding a dependency with `vgo`:
~~~
$ vgo get github.com/emicklei/go-restful@v2.7.0 
$ vgo list -m
github.com/cznic/fileutil             v0.0.0-20180108211300-6a051e75936f
github.com/emicklei/go-restful        v0.0.0-20180416204930-2810ccc68e0c
github.com/go-stack/stack             v1.7.0
...
~~~
as you can see, `go-restful` is added as an unversioned dependency (0.0.0) with the current git-SHA (2810ccc68e0c). well, this works, but is is not the desired behaviour (at least for me).

It would be great to support `vgo` in the long run. at the moment i have no problem with the `0.0.0` dependency but it would be great to support `vgo` in some later versions.

emicklei:
[...]

PR #413, opened 2019-09-05
feat: support google custome verb

ShiningRush:
support google [custome verb](https://cloud.google.com/apis/design/custom_methods)
resolve this issue. #394

### DIFF
--- a/curly.go
+++ b/curly.go
@@ -47,7 +47,7 @@ func (c CurlyRouter) SelectRoute(
 func (c CurlyRouter) selectRoutes(ws *WebService, requestTokens []string) sortableCurlyRoutes {
 	candidates := make(sortableCurlyRoutes, 0, 8)
 	for _, each := range ws.routes {
-		matches, paramCount, staticCount := c.matchesRouteByPathTokens(each.pathParts, requestTokens)
+		matches, paramCount, staticCount := c.matchesRouteByPathTokens(each.pathParts, requestTokens, each.hasCustomVerb)
 		if matches {
 			candidates.add(curlyRoute{each, paramCount, staticCount}) // TODO make sure Routes() return pointers?
 		}
@@ -57,7 +57,7 @@ func (c CurlyRouter) selectRoutes(ws *WebService, requestTokens []string) sortab
 }
 
 // matchesRouteByPathTokens computes whether it matches, howmany parameters do match and what the number of static path elements are.
-func (c CurlyRouter) matchesRouteByPathTokens(routeTokens, requestTokens []string) (matches bool, paramCount int, staticCount int) {
+func (c CurlyRouter) matchesRouteByPathTokens(routeTokens, requestTokens []string, hasCustomVerb bool) (matches bool, paramCount int, staticCount int) {
 	if len(routeTokens) < len(requestTokens) {
 		// proceed in matching only if last routeToken is wildcard
 		count := len(routeTokens)
@@ -72,7 +72,7 @@ func (c CurlyRouter) matchesRouteByPathTokens(routeTokens, requestTokens []strin
 			return false, 0, 0
 		}
 		requestToken := requestTokens[i]
-		if hasCustomVerb(routeToken){
+		if hasCustomVerb{
 			if !isMatchCustomVerb(routeToken, requestToken) {
 				return false, 0, 0
 			}
--- a/curly.go
+++ b/curly.go
@@ -57,7 +57,7 @@ func (c CurlyRouter) selectRoutes(ws *WebService, requestTokens []string) sortab
 }
 
 // matchesRouteByPathTokens computes whether it matches, howmany parameters do match and what the number of static path elements are.
-func (c CurlyRouter) matchesRouteByPathTokens(routeTokens, requestTokens []string, hasCustomVerb bool) (matches bool, paramCount int, staticCount int) {
+func (c CurlyRouter) matchesRouteByPathTokens(routeTokens, requestTokens []string, routeHasCustomVerb bool) (matches bool, paramCount int, staticCount int) {
 	if len(routeTokens) < len(requestTokens) {
 		// proceed in matching only if last routeToken is wildcard
 		count := len(routeTokens)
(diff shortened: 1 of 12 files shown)


ShiningRush:
resolve this issue. #394

custom_verb_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func TestRemoveCustomVerb(t *testing.T) {
+	testCase := []struct{
+		path string
+		expectedPath string
+	}{
+		{"/{userId}:init", "/{userId}"},
+		{"/{userId:init}", "/{userId:init}"},
+		{"/users/{id:init}:init", "/users/{id:init}"},
+		{"/users/{id}", "/users/{id}"},
```
emicklei: this will fail:   "/init/users/{id:init}:init" , "/init/users/{id:init}"

custom_verb.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func isMatchCustomVerb(routeToken string, pathToken string) bool {
+	rs := customVerbReg.FindStringSubmatch(routeToken)
+	if len(rs) < 2 {
+		return false
+	}
+
+	customVerb := rs[1]
+	specificVerbReg := regexp.MustCompile(fmt.Sprintf(":%v$", customVerb))
```
emicklei: better use %s instead of %v

curly.go
```diff
@@ -72,6 +72,15 @@ func (c CurlyRouter) matchesRouteByPathTokens(routeTokens, requestTokens []strin
 			return false, 0, 0
 		}
 		requestToken := requestTokens[i]
+		if hasCustomVerb(routeToken){
```
emicklei: the hasCustomVerb is doing a regexp operation. this will affect performance. for this reason, I would like to introduce a field in the Router called   CustomVerbEnabled with an default value of false.

custom_verb_test.go
ShiningRush: i added it to case, and run success,  is something wrong?
```
	testCase := []struct{
		path string
		expectedPath string
	}{
		{"/{userId}:init", "/{userId}"},
		{"/{userId:init}", "/{userId:init}"},
		{"/users/{id:init}:init", "/users/{id:init}"},
		{"/users/{id}", "/users/{id}"},
		{"/init/users/{id:init}:init" , "/init/users/{id:init}"},
	}
```

emicklei: tbh, i did not execute this test. i should have said "will this fail?".  I came up with this test because I saw "ReplaceAllString"  in removeCustomVerb.   And this case "/:init/users:init/" is probably not a valid path, right?

ShiningRush: emmm, i think it likes "/users/{id:init}:init", so i think the case is not necessary.
and replace regex is ":customVerb$", it will just replace last word.

path_processor.go
```diff
@@ -29,6 +29,11 @@ func (d defaultPathProcessor) ExtractParameters(r *Route, _ *WebService, urlPath
 		} else {
 			value = urlParts[i]
 		}
+		if hasCustomVerb(key) {
```
ShiningRush: is here need to check CustomVerbEnabled?

custom_verb_test.go
emicklei: ok, good!

path_processor.go
emicklei: yes, do you agree?

ShiningRush: yes, i should be checked

curly.go
ShiningRush: now it will calculate a route has a custom verb when build route

route.go
```diff
@@ -49,11 +49,15 @@ type Route struct {
...
 	contentEncodingEnabled *bool
+
+	// indicate route path has custom verb
+	hasCustomVerb bool
 }
 
 // Initialize for Route
 func (r *Route) postBuild() {
 	r.pathParts = tokenizePath(r.Path)
+	r.hasCustomVerb = hasCustomVerb(r.Path)
```
emicklei: nice!

curly.go
emicklei: so, we don't need the CustomVerbEnabled !

emicklei:
to complete the PR, we should have an entry in the features list on the README refering to https://cloud.google.com/apis/design/custom_methods and a minimal example in the /examples folder.

ShiningRush:
@emicklei , i have added a example (examples/restful-google-custom-method.go) and updated readme.md,

examples/restful-google-custom-method.go
```diff
@@ -0,0 +1,27 @@
...
+	. "github.com/emicklei/go-restful"
+	"io"
+	"log"
+	"net/http"
+)
+
+// This example shows how to create a Route with google custom method
+// Requires the use of a CurlyRouter and path should end with the custom method
+//
+// GET http://localhost:8080/resource/some-resource-id:init
```
emicklei: resource -> basepath

examples/restful-google-custom-method.go
```diff
@@ -0,0 +1,27 @@
...
+
+func main() {
+	DefaultContainer.Router(CurlyRouter{})
+	ws := new(WebService)
+	ws.Route(ws.GET("/basepath/{resourceId}:init").To(fromPathParam))
```
emicklei: i think it is good to add another Route to demonstrate that the verb is causing another function to be called.

so next to :init have :reset or some other verb you can think of.

ShiningRush: @emicklei , well, i have already fixed example

emicklei:
i think we are ready to merge

ShiningRush:
ok, lets do it

ShiningRush:
i found a issue #374 , and in this commit i added a `go.mod` file to support `go mod `,  is here would be any problems? @emicklei

emicklei:
well module support was quite a challenge with subpackages not importing a version. I will have a look again.  Would be better to separate that from this PR.

ShiningRush:
@emicklei ok, i removed the `go.mod` file, but it is worth to say that `go mod replace` can not import a local package without `go.mod`, this will cause some inconvenience to `go-restful`s contributors when they are debugging in local environment.

emicklei:
yes, i understand. It will be one of the first changes next to do.

ShiningRush:
good!

PR #414, opened 2019-10-10
handle path params with prefixes and suffixes

jaswdr:
In current implementation if you have a path parameter with some prefix or suffix the path processor doesn't extract the values correctly, see some examples:

```
/foo_{var}
/{var}_foo
/foo_{var}_bar
```

This patch handle those cases using regexp to correctly extract values.

### DIFF
--- a/path_processor.go
+++ b/path_processor.go
@@ -2,6 +2,7 @@ package restful
 
 import (
 	"bytes"
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


emicklei:
i am worried about the performance impact of this change. Creating and executing regexp can be expensive.

jaswdr:
@emicklei ack, I used now strings indexes like the previous code, please let me now if this feel better

jaswdr:
ping @emicklei

emicklei:
thank you for contributing

PR #417, opened 2019-11-27
add Go module support starting with v3

emicklei:


emicklei:
I am closing this as the branch will be kept alive for as long as v3 exists, forever?

Issue #423, opened 2020-02-03
Can it support S3 client?


sfzeng:
Hi, I'm trying to using it to implement an AWS S3 compatible s3 server, and I tried to use s3 client to test it, but I got '415: Unsupported Media Type' error when creating a bucket and initing multipart upload. For AWS S3, it's valid to do these two requests without a request body, that means it's valid to do PUT/POST request without a request body. So do you have any suggestions? Can I add PUT&POST to line 112 of route.go as below?
![20200203112448](https://user-images.githubusercontent.com/38932432/73623422-e083ad80-4677-11ea-8bec-8b666e6e9dc4.png)

P.S. I'm using go_s3tests: https://github.com/ceph/go_s3tests/#run-the-tests.

Thanks !!

emicklei:
thank you for reporting this. I will have a look at your usecase and come back with a suggestion asap

emicklei:
I propose to make a change to Route such that you can override this list of methods on a per Route basis. For that I am thinking about introducing a method to change the field allowedMethodsWithoutContentType which is default set to the methods currently present at line 112

emicklei:
have a look at https://github.com/emicklei/go-restful/pull/424

sfzeng:
Hi @emicklei , thank you very much, that works for me now !

emicklei:
can this be closed?

PR #424, opened 2020-02-05
feature: override list of method allowed without content-type ( issue #423 )

emicklei:


sfzeng:
Thanks, that works well for me.

PR #426, opened 2020-02-24
support describing response headers

### LINKED ISSUE #419: can not specify response header?
```go
// Returns allows you to document what responses (errors or regular) can be expected.
// The model parameter is optional ; either pass a struct instance or use nil if not applicable.
func (b *RouteBuilder) Returns(code int, message string, model interface{}) *RouteBuilder {
	err := ResponseError{
		Code:      code,
		Message:   message,
		Model:     model,
		IsDefault: false, // this field is deprecated, use default response instead.
	}
	// lazy init because there is no NewRouteBuilder (yet)
	if b.errorMap == nil {
		b.errorMap = map[int]ResponseError{}
	}
	b.errorMap[code] = err
	return b
}
```

now I can not use code to describe the response header, it can not automatically generate a complete swagger doc.

is there any way I can generate response header spec?

zhulijian1:
Fixes #419 

In the OpenAPI Specification, response object has a headers field. (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#responseObject)

But there is no field about headers in struct ResponseError. So I can not use code to describe the response header.

```go
type ResponseError struct {
	Code      int
	Message   string
	Model     interface{}
	IsDefault bool
}
```
I added the Headers field with reference to the OpenAPI Specification. And in order to maintain compatibility, I did not modify Returns method, but added a new method ReturnsWithHeaders.

### DIFF
--- a/route_builder.go
+++ b/route_builder.go
@@ -176,6 +176,15 @@ func (b *RouteBuilder) Returns(code int, message string, model interface{}) *Rou
 	return b
 }
 
+// ReturnsWithHeaders is similar to Returns, but can specify response headers
+func (b *RouteBuilder) ReturnsWithHeaders(code int, message string, model interface{}, headers map[string]Header) *RouteBuilder {
+	b.Returns(code, message, model)
+	err := b.errorMap[code]
+	err.Headers = headers
+	b.errorMap[code] = err
+	return b
+}
+
 // DefaultReturns is a special Returns call that sets the default of the response.
 func (b *RouteBuilder) DefaultReturns(message string, model interface{}) *RouteBuilder {
 	b.defaultResponse = &ResponseError{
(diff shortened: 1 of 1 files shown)


emicklei:
your change looks good to me.
unfortunately, the OpenAPI specifics are now in the go-restful package ; most of it I was able to put in the go-restful-openapi package.
However, I can imagine that the non-OpenAPI counterpart will look just like you included in the PR.
I does not make sense to come up with a different type (another name) with will have the same fields anyway.

zhulijian1:
My aim is to achieve the ability to define response headers. But if I don't add that structure, I can not find any structure suitable for describing the response headers. 
Do you have any good suggestions on what I should do to achieve my aim?

emicklei:
@zhulijian1 no i do not. I will merge this change and make it available in v2 and v3

zhulijian1:
thanks :)

PR #437, opened 2020-06-21
List available representations in 406 body

### LINKED ISSUE #434: Acceptable representations not listed in body of '406: Not Acceptable' responses
RFC 7231 says that the server *should* include a list of representations in the body of a 406 response: https://tools.ietf.org/html/rfc7231#section-6.5.6

Currently restful just says:
```
HTTP/1.1 406 Not Acceptable
Date: Sun, 24 May 2020 14:31:55 GMT
Content-Length: 19
Content-Type: text/plain; charset=utf-8

406: Not Acceptable
```
<details>
<summary>Test server</summary>

```
package main

import (
	"io"
	"log"
	"net/http"

	"github.com/emicklei/go-restful"
)

func main() {
	ws := new(restful.WebService)
	ws.Route(ws.GET("/hello").Produces("text/plain").To(hello))
	restful.Add(ws)
	log.Fatal(http.ListenAndServe(":8080", nil))
}

func hello(req *restful.Request, resp *restful.Response) {
	io.WriteString(resp, "world")
}

```
</details>

But the RFC doesn't have an opinion on what form that list should take. As restful already provides a text/plain body, I think it should be expended to include each of the route's produces, perhaps:
```
406: Not Acceptable

Acceptable: text/plain, application/json
```

If you agree with the above, I'm happy to do the work.

brackendawson:
Becasue RFC7231 says we should.

fixes #434

### DIFF
--- a/jsr311.go
+++ b/jsr311.go
@@ -147,7 +147,14 @@ func (r RouterJSR311) detectRoute(routes []Route, httpRequest *http.Request) (*R
 		if trace {
 			traceLogger.Printf("no Route found (from %d) that matches HTTP Accept: %s\n", len(previous), accept)
 		}
-		return nil, NewError(http.StatusNotAcceptable, "406: Not Acceptable")
+		available := []string{}
+		for _, candidate := range previous {
+			available = append(available, candidate.Produces...)
+		}
+		return nil, NewError(
+			http.StatusNotAcceptable,
+			fmt.Sprintf("406: Not Acceptable\n\nAvailable representations: %s", strings.Join(available, ", ")),
+		)
 	}
 	// return r.bestMatchByMedia(outputMediaOk, contentType, accept), nil
 	return candidates[0], nil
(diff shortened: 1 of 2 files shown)


brackendawson:
I chose to not remove duplicates from `available` because this is the last thing in `detectRoute()` and therefore there should be none?

brackendawson:
Also was this supposed to test a 406 response? https://github.com/emicklei/go-restful/blob/8487208e17d772637e9b2c61c522427fdd2fe83d/web_service_test.go#L154

emicklei:
> I chose to not remove duplicates from `available` because this is the last thing in `detectRoute()` and therefore there should be none?

please explain why not removing dups?

brackendawson:
> please explain why not removing dups?

Because the Accept header is the last thing checked in the router, a valid path should not contain any duplicate media types. If it did then a successful call would fail.

It does look like the sort of thing that would get missed if another switch was added to the router after Accept. I can add deduplication if you like? It is only exposed in the error case.

PR #439, opened 2020-07-12
bufix

zhangweidongyyl:
【bugfix】fix createUser and updateUser bug in example


emicklei:
i do not think this is a but.  REST says that to create a resource you either use a POST on a collection, or a PUT or an URI

PR #446, opened 2020-08-19
Enable content encoding on Handle and ServeHTTP

caskolkm:


emicklei:
thank you for contributing. I will create a new tag both on master and v3

PR #451, opened 2020-10-09
make read request body many times more easier.

xdhuxc:


emicklei:
thank you for your contribution.

this change could have an unforeseen impact to other users as it introduces more computation and memory.
also, I think adding this re-reading of requests to the framework only supports a very limited set of use cases.

xdhuxc:
@emicklei yes, you are right. I am really sorry that I was not thoughtful enough. but can we provide another method , with remaining the origin ReadEntity method, to support this? My case is that generating audit and event sdk for some system, this will make my code more elegant.

emicklei:
something like ReadCachedEntity?

xdhuxc:
@emicklei Yes, It will have no effect on the existed project and provider method to remain original http request Body to some use cases.

xdhuxc:
It can be ok.

Issue #449, opened 2020-10-09
HTTP response body was broken, if struct to be converted to JSON has boolean value.


shu-mutou:
v2.13.0 goes well, but 2.14.0 or later version occurs this issue.

emicklei:
thank you for reporting this. I will investigate what is going on.

emicklei:
did you try v2.14.2  as it has fixed w.r.t 2.14.0

shu-mutou:
I'd tried 2.14.0 and 2.14.2

emicklei:
thanks, do you have a test that shows this problem?

shu-mutou:
There are no results such as unit tests, but I suspect it is due to the use of the `rune` function.

shu-mutou:
Additional info: I found that changing `EnableContentEncoding` from `true` to `false` with 2.14.2 works well.

shu-mutou:
I tried v3, v3.2.0 works but v3.3.1 does not.

emicklei:
@shu-mutou does it fail for the same reason?

shu-mutou:
@emicklei Yes. Completely same.

emicklei:
@shu-mutou  can you have a look at the example i just added (v3 branch).

With content encoding enabled + accept header , you can gzipped content of the users list.

shu-mutou:
@emicklei thank you for the example!

In our implementation, I found the responses were gzipped twice.
And I commented out following, then it seems to work fine for our implementation.
https://github.com/emicklei/go-restful/blob/master/container.go#L240-L257

I guess check like following is needed for `dispatch` in `container.go` like in `ServeHTTP` or `Handle`.
https://github.com/emicklei/go-restful/blob/master/container.go#L305-L309

emicklei:
thank you for this addition insight. will get to it now

shu-mutou:
We confirmed that v3.3.3 works fine!! Thanks.

PR #453, opened 2020-10-22
fix for issue #449, duplicate compressing

emicklei:


shu-mutou:
Thank you very much!! This works well for us! 😄

Issue #459, opened 2020-12-07
Proposal: Add Route metadata to request attributes or allow adding attributes to routes


Fapiko:
It would be nice to see metadata or some way to inject attributes of routes into the request object so they are accessible in handlers and filters. It seems like it would be fairly easy to do it when the request is wrapped: https://github.com/emicklei/go-restful/blob/80bbcf816141a5df050090c4914ad294ede63374/route.go#L69-L77 It already has an `Attributes` field but those seem to be scoped to an inbound request so I'm not sure how that field usage is intended to differ from `Context`

This would be incredibly handy for handling things like authorization. In my case I'd like to tag each of the routes with an `object` and `permission` required to access that resource, then add a filter to verify the authenticated client has authorization to that object and permission. 

It would allow something like this when building the routes:
```
ws.Route(ws.GET("/users").To(u.getUsers).
	Doc("Gets a list of all users in the platform").
	Metadata(restfulspec.KeyOpenAPITags, tags).
	Attribute("auth_object", "users").
	Attribute("auth_permission", "view").
	Writes([]UserResponse{}))
```

I was attempting to use the `metadata` field but as it stands right now I cannot figure out a way to link a request from the handler or a filter back to a route to look that info up.

emicklei:
interesting. 

Yes, the attributes field was introduced at a time that Context was not available or not part of the standard pkgs.
I will have a look if and how attributes can be made available to the request.

emicklei:
have a look at https://github.com/emicklei/go-restful/pull/462

emicklei:
@Fapiko does this help your usecase?

emicklei:
this feature is include in v.3.5.1

PR #462, opened 2020-12-13
add access to Route from Request, issue #459

emicklei:


Fapiko:
Apologies - didn't see your previous conversation moment. I think this would work perfectly!

Issue #467, opened 2021-04-12
Request.SelectedRoutePath panics if no route is selected


axw:
`Request.SelectedRoutePath` panics when called on a request with selected route. For example, this happens when using Elastic APM's `apmrestful` filter: https://github.com/elastic/apm-agent-go/blob/f62afdfda01d0cbe73f477c357d5b8a3017e5144/module/apmrestful/filter.go#L55-L66

This is caused by https://github.com/emicklei/go-restful/pull/462/files. Previously the method would return "".

emicklei:
thank you for reporting this. somehow the tests did not capture this. will work on a fix asap

emicklei:
https://github.com/emicklei/go-restful/pull/468

PR #468, opened 2021-04-12
fix handling no match access selected path #467

emicklei:


axw (approved):
thanks :)

Request has a selectedRoutePath field which I believe is no longer referenced, looks like leftovers from the recent change.

PR #472, opened 2021-07-12
Issue 471: Fix panic when removing a route that doesn't exist in w.routes.

jeffreydwalter:
This PR fixes the remaining panic in `RemoveRoute()`.

### DIFF
--- a/web_service.go
+++ b/web_service.go
@@ -181,12 +181,12 @@ func (w *WebService) RemoveRoute(path, method string) error {
     }
     w.routesLock.Lock()
     defer w.routesLock.Unlock()
-    newRoutes := make([]Route, 0)
-    for ix := range w.routes {
-        if w.routes[ix].Method == method && w.routes[ix].Path == path {
+    newRoutes := []Route{}
+    for _, route := range w.routes {
+        if route.Method == method && route.Path == path {
             continue
         }
-        newRoutes = append(newRoutes, w.routes[ix])
+        newRoutes = append(newRoutes, route)
     }
     w.routes = newRoutes
     return nil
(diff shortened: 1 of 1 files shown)


web_service.go
```diff
@@ -176,22 +176,20 @@ func (w *WebService) Route(builder *RouteBuilder) *WebService {
...
-		current++
-	}
-	w.routes = newRoutes
-	return nil
+    if !w.dynamicRoutes {
+        return errors.New("dynamic routes are not enabled.")
+    }
+    w.routesLock.Lock()
+    defer w.routesLock.Unlock()
+    newRoutes := make([]Route, 0)
```
emicklei, line 184: now that the length is not of interest, we can write []Route{}.

web_service.go
```diff
@@ -176,22 +176,20 @@ func (w *WebService) Route(builder *RouteBuilder) *WebService {
...
+    for ix := range w.routes {
```
emicklei, line 185: now that the index is not needed, we can write    for _ , each := range w.routes {   if each.Method == method.... }
so using `each` for `w.routes[ix]`

jeffreydwalter:
@emicklei I made the changes you suggested. Would like to get this PR merged so I can vendor it back into my codebase. Please let me know if there's anything else I can do to facilitate. Thanks!

jeffreydwalter:
@emicklei thinking about this further, what do you think about setting the capacity of `newRoutes`, like this: `newRoutes := make([]Route, 0, len(w.routes))`? This would potentially allocate an extra item in the slice in the case the route is found and removed, but would cut down on slice allocations. If you think that is better, I'm happy to make another PR.

jeffreydwalter:
@emicklei what do you say in regard to my last comment? ☝🏻

emicklei:
@jeffreydwalter , removing a Route is not on the hot-path so any optimisations here are not significant.

PR #477, opened 2021-09-13
feat(extensions): adds vendor extensions to all entities

erraggy:
Swagger provides vendor extensions that allow custom properties to be added to any of the swagger entities (params, operations, responses, etc...). This adds a new companion type: `restful.ExtensionProps` that provides a `map[string]interface{}` to store extension values by their string key and the `AddExtension(key string, value interface{})` method to store values.

Since swagger is handled by the `go-restful-openapi` module, the changes here are required in its own pull request: emicklei/go-restful-openapi#71

For more details on swagger extensions, please see: https://swagger.io/docs/specification/2-0/swagger-extensions/


go.mod
```diff
@@ -1,3 +1,12 @@
 module github.com/emicklei/go-restful/v3
 
 go 1.13
+
```
emicklei, line 4: These dependencies should not be required. Most likely they are added because of the examples. I will revisit this.

emicklei: please remove the go.mod and go.sum from the PR

erraggy: backed those out with: e2c750eb4b5f3a84e37508c7919eb9ed847b794c

emicklei:
Thank you for contributing the making this last change. One last preference is to not use an abbreviation for ExtensionProperties. As you can see in the package, no other type name is shortened.

erraggy:
> Thank you for contributing the making this last change. One last preference is to not use an abbreviation for ExtensionProperties. As you can see in the package, no other type name is shortened.

Changes applied. Thanks for the prompt reviews 👍🏻

PR #491, opened 2022-04-05
[Minor] Bad practice to have a mix of Receiver types.

MushuEE:
I was using an old version and saw that 
```
r.contentEncodingEnabled = &enabled
```
Fix was added, however String() should be updated to also use a pointer receiver.

Per Golang Docs "In general, all methods on a given type should have either value or pointer receivers, but not a mixture of both."

### DIFF
--- a/route.go
+++ b/route.go
@@ -168,7 +168,7 @@ func tokenizePath(path string) []string {
 }
 
 // for debugging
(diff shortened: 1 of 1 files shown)


emicklei:
thank you for contributing

Issue #489, opened 2022-03-29
[security] Authorization Bypass Through User-Controlled Key


JamieSlome:
@emicklei - following on from #488 and https://huntr.dev/bounties/be837427-415c-4d8c-808b-62ce20aa84f1/, we are sharing the details of the report as requested in the `SECURITY.md`.

## Description
Hello go restful maintainer team, I would like to report a security concerning your CORS Filter feature.

Go restful allows user to specify a CORS Filter with a configurable AllowedDomains param - which is an array of domains allowed in CORS policy.

However, although there's is already another param called allowedOriginPatterns used for matching origin using regular expression, all domains in AllowedDomains is also used as regular expression to check for matching origin in this code in file cors_filter.go:

if len(c.allowedOriginPatterns) == 0 {
            // compile allowed domains to allowed origin patterns
            allowedOriginRegexps, err := compileRegexps(c.AllowedDomains)
            if err != nil {
                return false
            }
            c.allowedOriginPatterns = allowedOriginRegexps
        }

        for _, pattern := range c.allowedOriginPatterns {
            if allowed = pattern.MatchString(origin); allowed {
                break
            }
        }
So by this, if the user input example.com to be one of domain in AllowedDomains, all domains starting with example.com would be acceptable.

## Proof of Concept
Install go restful and create a file main.go with this content:
package main

import (
    restful "github.com/emicklei/go-restful/v3"
    "io"
    "net/http"
)

func main() {
    container := restful.NewContainer()

    ws := new(restful.WebService)
    ws.Route(ws.GET("hello").To(hello))
    container.Add(ws)
    server := &http.Server{Addr: ":8000", Handler: container}

    //container.Filter(logHeaders)
    cors := restful.CrossOriginResourceSharing{
        ExposeHeaders: []string{"X-My-Header"},
        AllowedDomains: []string{"example.com"},
        CookiesAllowed: true,
        Container: container,
    }
    container.Filter(cors.Filter)
    server.ListenAndServe()
}

func hello(req *restful.Request, resp *restful.Response) {
    io.WriteString(resp, "world")
}

In the above code, example.com is configured as an allowed domain.

Run the above code and access link /hello with Origin Header = example.com.hacker.domain and see that the request gets through CORS policy and response looks like this:
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: example.com.hacker.domain
Access-Control-Expose-Headers: X-My-Header
[...]

PR #493, opened 2022-04-30
use exact matching of allowed domain entries, issue #489

emicklei:


crenshaw-dev:
The docs should be updated to indicate that each allowed domain isn't a _potential_ regexp pattern, it _is_ a regexp pattern. So `^some.example.com$` matches `somexexample.com`. Even after the PR, `^some\.example\.com` matches `some.example.com.org`.

emicklei:
@JamieSlome is your hunt issue resolved by this PR?

cors_filter.go
```diff
@@ -21,7 +21,11 @@ import (
 type CrossOriginResourceSharing struct {
 	ExposeHeaders  []string // list of Header names
 	AllowedHeaders []string // list of Header names
-	AllowedDomains []string // list of allowed values for Http Origin. An allowed value can be a regular expression to support subdomain matching. If empty all are allowed.
+	// AllowedDomains list of allowed values for Http Origin.
+	// An allowed value can be a regular expression to support subdomain matching.
+	// Non-regular expression values will be changed into an exact match: ^yourdomain.com$
```
crenshaw-dev, line 26: I'm not sure `^yourdomain.com$` is an intuitive "exact match". The `.` is still a wildcard. You'd have to do `fmt.Sprintf("^%s$", regexp.QuoteMeta(each))` to get an "exact match".

crenshaw-dev, line 26: It's not clear to me from this what qualifies as "non-regular expression". `example.com` and `.*` and `example\.com` are all regular expressions, even though they don't start with `^` or end with `$`.

JamieSlome:
@emicklei:

Just (cc) @sickcodes and @ktg9 to see if they think this addresses the reported issue 👍

cors_filter.go
steven-collins-omega: agreed. `^www.amazon.com$` still matches `wwwqamazon.com` and `www.amazonqcom`.

steven-collins-omega:
@emicklei if I'm reading the code correctly, this just simply isn't going to work in practice in most cases. `isOriginAllowed()` is called on the literal value of the Origin header, [which looks something like](https://html.spec.whatwg.org/multipage/origin.html#ascii-serialisation-of-an-origin) `https://some.example.com:8080`. Technically the documentation does say that `AllowedDomains` is a "list of allowed values for Http Origin", which is what it is. However, given the name `AllowedDomains` and the `cors_filter` test code, a library user would be likely to interpret it as a list of allowed domains. Such a user will find, after this proposed change, that nothing is allowed anymore. `^some.example.com$` doesn't match `https://some.example.com:8080`.

emicklei:
@steven-collins-omega thank you for commenting on this. I agree that it is not correct and confusing. I will give it second time to come up with a solution

emicklei:
hi all, I thinking about changing the behaviour of CrossOriginResourceSharing:

- AllowedDomains will have entries that must include the origin value from the request by string comparing.
- add the AllowedDomainFunc field which can be set with a function takes takes the `origin` and returns a bool for a match.  This allows the developer define its own logic.

cors_filter.go
```diff
@@ -120,35 +133,22 @@ func (c CrossOriginResourceSharing) isOriginAllowed(origin string) bool {
...
 		return true
 	}
 
-	allowed := false
+	// exact match on each allowed domain
 	for _, domain := range c.AllowedDomains {
-		if domain == origin {
-			allowed = true
-			break
+		if domain == ".*" || domain == origin {
```
crenshaw-dev, line 144: Should `domain` and `origin` be lowercased before comparison?

emicklei: yes, i think this should happen.  RFC4343: Domain Name System (DNS) names are "case insensitive".

PR #501, opened 2022-06-30
Master

armersong:
support multipart/form-data


emicklei:
hi, thank you for contributing.
I noticed that you added a go.mod file.
The master branch is for the non-Go module versions <v3.
The v3 branch is the current and I think you PR must target v3 instead.

Issue #497, opened 2022-05-30
[security] Path parser inconsistency could lead to bypass several security checks in emicklei/go-restful


JamieSlome:
Reported via huntr.dev [here](https://huntr.dev/bounties/1beeab79-9a3a-4674-bd0a-78fff7cd6cca/).

Created: `May 28th 2022`

## Description

There is a inconsistency between how golang(and other packages) and go-restful parses url path. This incosistency could lead several security check bypass in a complex system.

## Steps to reproduce

Copy and run the code below
```
package main

import (
    "fmt"
    "html"
    "log"
    "net/http"
)

func main() {

    http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
        fmt.Fprintf(w, "Hello, %q", html.EscapeString(r.URL.Path))
    })

// If a user request matches with this path then it will be checked and if it didn't match then the request will be directly forwarded to the go-restful server.
    http.HandleFunc("/users/id_will_be_here", func(w http.ResponseWriter, r *http.Request) {
        fmt.Fprintf(w, "I'm checking if the user is authorized to see this user info. If he is authorized then I will forward this request to the go-restful server which will return the user info")
    })

    log.Fatal(http.ListenAndServe(":8081", nil))

}
```
Now If you send a request to the http://localhost:8081/users/id_will_be_here then you will get a hit to the expected endpoint and it will check the authorization part
Now if you send a request to the http://localhost:8081/users/id_will_be_here/ (notice the extra / in last) then this server won't process this request as the path name doesn't match but the problem is that go-restful treat this path and the previous as same.
So this inconsistency could lead some issues in this scenerio as the first sever wont check the security checks but the second server(go-restful) will return user data.

## Impact
Security check bypass

## Occurrences

[web_service.go L80](https://github.com/emicklei/go-restful/blob/7c971ca4d0fd3dc297bb263f60729bcfda65853b/web_service.go#L80)

emicklei:
Thank you for reporting this. I will investigate this further.

emicklei:
i was able to reproduce this issue.

your example output

```
~/tmp/issue497
$ curl http://localhost:8081/users/id_will_be_here
I'm checking if the user is authorized to see this user info. If he is authorized then I will forward this request to the go-restful server which will return the user info

emicklei@Ernests-iMac-10 on Sat Jun 04 at 10:35 AM
~/tmp/issue497
$ curl http://localhost:8081/users/id_will_be_here/
[...]

PR #511, opened 2022-09-30
changed tokenizer to match std route match behavior; do not trimright… issue #497

emicklei:
… the path


starzhanganz:
@emicklei is that possible to get this merged? thanks.

Issue #514, opened 2022-10-12
Support for multiple return types


Jefftree:
Is there any plan to support an endpoint capable of returning multiple return types? Eg: a `ws.Route(...).Writes(TypeA{}}, TypeB{})`

emicklei:
can you explain the usecase?  does the OpenAPI also support this?

Jefftree:
OpenAPI V3 supports returning a oneOf (https://swagger.io/specification/#discriminator-object). Kubernetes has endpoints that could return one of multiple values and I think it'd be useful if the WebService object could express that as well.

emicklei:
@Jefftree , did you have a look at the PR?

Jefftree:
Thanks for the addition, will test it out this week but marking this as closed for now.

PR #515, opened 2022-10-19
allow multiple samples for Write, issue #514

emicklei:


SVilgelm:
same about readSamples :) 
we have bunch of endpoints that can work with the different types, like a general create api where is a type of object is a field in payload :)))

emicklei:
> same about readSamples :)
> we have bunch of endpoints that can work with the different types, like a general create api where is a type of object is a field in payload :)))

that should be an easy addition to this PR.

PR #517, opened 2022-11-07
Add some kinds of log

Feizekai:


emicklei:
This will break compatibility for users that implement their own Logger interface. Sorry, cannot merge this

PR #518, opened 2022-11-09
chore: example handling request parameters with httpin

ggicci:


emicklei:
hi, your contribution clearly is a promotion of your package and demonstration how to integrate with go-restful. A few other examples have the same goal, for different packages of course. 

I am willing to accept your PR with the disclaimer that you yourself are responsible for maintaining the example. If at some point in time the example is broken of some period of time then it may be removed from the repository.

ggicci:
thanks @emicklei for the clarification, i will keep the example updated :)

PR #539, opened 2024-01-03
remove the dependency on github.com/json-iterator/go.

dominiquelefevre:
That module uses github.com/modern-go/reflect2 which is broken.

First, it is brittle because it relies on golang's implementation details that are not stable across golang releases. For example, reflect2 was broken when go 1.18 came out.

Second, reflect2 is effectively unmaintained:
  https://github.com/modern-go/reflect2/issues/24#issuecomment-916998286

Json-iterator itself has had correctness issues like
  https://github.com/json-iterator/go/pull/413.

The Kubernetes project has mostly removed the dependency on json-iterator in patches like
  https://github.com/kubernetes/kubernetes/pull/105030.

Moreover, the Kubernetes authors found it out that json-iterator puts a lot of load on the allocator, so even the performance gains are questionable.

Let us remove dependencies on json-iterator and reflect2.


emicklei:
thank you for contributing to this package. I will review your changes.

emicklei (approved):

michaelbeaumont:
I think it was a mistake to remove the ability to customize `MarshalIndent, NewDecoder, NewEncoder` functions here and it makes it even more of a breaking change.

PR #540, opened 2024-01-09
fix: restore ability to customize JSON handling

michaelbeaumont:
These functions are no longer public after https://github.com/emicklei/go-restful/pull/539 which made it a breaking change and is also unnecessary given the goal of the PR.

### DIFF
--- a/entity_accessors.go
+++ b/entity_accessors.go
@@ -11,6 +11,12 @@ import (
 	"sync"
 )
 
+var (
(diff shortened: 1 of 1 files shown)


emicklei (approved):

emicklei:
thank you providing a fix ; clearly I missed the "breaking change" part when reviewing it

michaelbeaumont:
Thanks @emicklei!

PR #543, opened 2024-03-07
fix: Improper handling of empty POST requests

mayooot:
issue:  [#532](https://github.com/emicklei/go-restful/issues/532)


liggitt:
Is an explicitly zero-length body *always* invalid? If not, the merged PR for this looks incorrect at first glance...

Things like https://lists.w3.org/Archives/Public/ietf-http-wg/2010JulSep/0276.html led me to think a zero-length body on put/post/patch was valid

How else would you send a `text/plain` POST or PUT intended to set an empty value?

(we spotted this while reviewing an update of the library for Kubernetes in https://github.com/kubernetes/kubernetes/pull/128919)

emicklei:
hi @liggitt , thank you for sharing your thoughts. I will have a look at the docs suggested.

Issue #545, opened 2024-05-21
regex route path allows partial match


haitch:
to accept case-insensitive paths, we use `{h:(?i)hello}`, to accept /HELLO, /hello, /Hello.

but with this approach, it also triggered a bug where /xxxHello, /yyyHello also got accepted.

```golang
package main

import (
	"io"
	"log"
	"net/http"

	restful "github.com/emicklei/go-restful/v3"
)

func main() {
	ws := new(restful.WebService)
	ws.Route(ws.GET("/{h:(?i)hello}").To(hello))
	restful.Add(ws)

	log.Fatal(http.ListenAndServe(":8080", nil))
}

func hello(req *restful.Request, resp *restful.Response) {
	io.WriteString(resp, "world")
}
```

haitch:
After dig into the code, I think the key is `func (c CurlyRouter) regularMatchesPathToken(routeToken, colon, requestToken)`, where it try to match a requestToken `xxxhello` to a routeToken `{h:(?i)hello}`.

in my example setup, `c.regularMatchesPathToken("{h:(?i)hello}", 3, "xxxhello")` was invoked. inside the function, curly bracket was removed, then, it's matching "(?i)hello" to "xxxhello", it did match hello, despect the prefix xxx.
```golang 
helloRegex := regexp.MustCompile("(?i)hello")
matches := helloRegex.FindStringSubmatch("xxxhello")
log.Println(matches) // []string{"hello"}
```

haitch:
did find a workaround by using more strict regex `{h:(?i)^hello$}`, the ^$ would ensure whole token match.

haitch:
send PR with example on how to handle this issue safely #546

emicklei:
thank you for reporting this issue and investigating the problem. I will have a look at the PR.

haitch:
thanks for merging the example PR, hopefully that will help someone. 

As a library, I think we want to prevent caller from make mistakes and cause misroute, can we consider have regex to match the whole routeToken with next minor release where you can introduce break changes.

emicklei:
i think we can close this.  re-open if you think otherwise

PR #546, opened 2024-05-21
add path template case-sensitive example

haitch:
- add example on how to correctly handle case-sensitive in path template #545 
- also comment on wrap http.ListenAndServe with log.Fatal

### DIFF
--- a/examples/hello/restful-hello-world.go
+++ b/examples/hello/restful-hello-world.go
@@ -16,7 +16,10 @@ func main() {
 	ws := new(restful.WebService)
 	ws.Route(ws.GET("/hello").To(hello))
 	restful.Add(ws)
-	log.Fatal(http.ListenAndServe(":8080", nil))
+
+	// DO NOT wrap http.ListenAndServe with log.Fatal in production
+	// or you won't be able to drain in-flight request gracefully, even you handle sigterm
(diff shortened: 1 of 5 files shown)


examples/hello/restful-hello-world.go
```diff
@@ -16,7 +16,10 @@ func main() {
 	ws := new(restful.WebService)
 	ws.Route(ws.GET("/hello").To(hello))
 	restful.Add(ws)
-	log.Fatal(http.ListenAndServe(":8080", nil))
+
```
emicklei, line 19: good point!

emicklei (approved):

examples/hello/restful-hello-world.go
haitch: thanks, for anyone without context, here is more reading: https://pkg.go.dev/net/http#Server.Shutdown

> When Shutdown is called, [Serve](https://pkg.go.dev/net/http#Serve), [ListenAndServe](https://pkg.go.dev/net/http#ListenAndServe), and [ListenAndServeTLS](https://pkg.go.dev/net/http#ListenAndServeTLS) immediately return [ErrServerClosed](https://pkg.go.dev/net/http#ErrServerClosed). Make sure the program doesn't exit and waits instead for Shutdown to return.

PR #548, opened 2024-05-24
better naming for routing related variable

haitch:

PR #549, opened 2024-05-25
fix misroute when dealing multiple webservice with regex

haitch:
feel pretty confident on this change: we skip regex tokens when detectWebService, but we still do regex match for each token on detectRoutes.

so there is no down side by doing extra regex match on detectWebService, and increase the score.

https://github.com/emicklei/go-restful/pull/548


haitch:
I still need this one to be merged, so I can safely upgrade some legacy codebase to v3.

I did notice there is v4 branch, and I [tested with v4 branch](https://github.com/haitch/go-restful/tree/v4), this change are still compatible, and passing all unittest.

PR #550, opened 2024-05-27
golang workflows for restful

haitch:


emicklei (approved):

emicklei:
I just noticed that the tests are using Go 1.21. However, the go.mod of this package states go 1.13.

haitch:
1. it will still works, workflow version is higher than the go mod version, which should be fine.
2. more of a issue is do you want this module to remain on go 1.13, golang ships pretty fast, 1.13 is already out of support, checkout https://endoflife.date/go 
3. this go mod version doesn't bother me too much, not having test on PR does bother me.

emicklei:
> 1. it will still works, workflow version is higher than the go mod version, which should be fine.

Ok.

> 2. more of a issue is do you want this module to remain on go 1.13, golang ships pretty fast, 1.13 is already out of support, checkout https://endoflife.date/go

Didn't know about that. Then it is time to upgrade.

> 3. this go mod version doesn't bother me too much, not having test on PR does bother me.

Agree

PR #557, opened 2024-06-24
dependabot to group PRs

haitch:
so this should only generate 1 PR every week.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,14 +5,20 @@
 
 version: 2
 updates:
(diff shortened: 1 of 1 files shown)


haitch:
@emicklei this should reduce the burden of maintenance.

dependabot
- is helpful to keep go mod up to date.
- is also helpful to keep update github workflows, we only have go test workflow as of now.

emicklei:
why was the target branch removed?

haitch:
> why was the target branch removed?

apparently some copy-paste error, while lost tracking v3 is the default branch.

PR #588, opened 2025-08-10
  Add configurable regex caching for CurlyRouter path   tokens and custom verbs

wenhug:
Adds optional regex pattern caching to improve CurlyRouter performance for applications with frequent path matching operations

### DIFF
--- a/curly.go
+++ b/curly.go
@@ -27,6 +27,17 @@ func SetPathTokenCacheEnabled(enabled bool) {
 	pathTokenCacheEnabled = enabled
 }
 
+// getCachedRegexp retrieves a compiled regex from the cache if found and valid.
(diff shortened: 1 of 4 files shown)


emicklei:
thank you for contributing, I will have a look at the optimization

wenhug:
This is a serious performance issue. Without this PR, every request needs to compile regex dynamically,  which uses about 20% CPU on our service. Dynamic regex compilation also allocates much memory and creates pressure for GC. Please consider merging this to fix the performance problem

emicklei (approved):
thank you for improving the performance and providing good tests that coverages the execution paths.
I wil make an update to the README and CHANGES to mention this new feature. Most likely in `v4` I will make caching the default.

